PR #4, opened 2025-06-03
Update containerization apple main

katiewasnothere:


katiewasnothere:
Note: CI will not pass since the test images are not yet created

PR #9, opened 2025-06-04
Source code documentation updates

dcantah:
This change adds documentation to quite a few existing public types that didn't have a blurb before.

Additionally, this fixes a couple things that I think either didn't make sense when going to document them:
- Rename ConnectionStream to VsockConnectionStream. This type only functions for vsock connections.
- Deletes NsLock+Closure. This was not used anywhere.
- Rename ContainerizationOCI/Config.swift to ImageConfig.swift.


Sources/ContainerizationOS/Mount/Mount.swift
```diff
@@ -30,6 +30,8 @@ private let _umount = Glibc.umount2
 /// Technically, this would be fine in the Linux subdirectory as it's Linux specific for now, but that
```
egernst: Should these still be ///? where does this get published if so?

dcantah: Probably not, good catch

Sources/ContainerizationOS/Mount/Mount.swift
```diff
@@ -108,6 +110,17 @@ extension Mount {
         return false
     }
 
+    public func mount(root: String, createWithPerms: Int16? = nil) throws {
```
crosbymichael, line 115: should we have a comment on this public method?

Sources/ContainerizationOS/Mount/Mount.swift
crosbymichael: Makes sense to remove

Sources/ContainerizationOS/Mount/Mount.swift
dcantah: Yes

Sources/cctl/ContainerStore.swift
```diff
@@ -35,7 +35,7 @@ extension String {
 
 #endif
 
-public final class ContainerStore: Sendable {
+public struct ContainerStore: Sendable {
```
katiewasnothere: Should this have a comment as well?

dcantah: I wasn't going to document the cctl surface, truthfully this shouldn't even be public as it's an executable. @crosbymichael that okay if I make this package/internal?

Sources/ContainerizationOS/AsyncSignalHandler.swift
```diff
@@ -17,6 +17,8 @@
 import Foundation
 import Synchronization
 
+/// Async friendly wrapper around DispatchSourceSignal. Provided an AsyncStream
+/// interface to get notified of received signals.
```
katiewasnothere: `Provides`?

Sources/ContainerizationOS/Socket/Socket.swift
```diff
@@ -44,6 +44,7 @@ let sysConnect = connect
 let sysIoctl: @convention(c) (CInt, CUnsignedLong, UnsafeMutableRawPointer) -> CInt = ioctl
 #endif
 
+/// Threadsafe socket wrapper.
```
katiewasnothere: nit: `thread-safe`

dcantah:
@crosbymichael ptal again

PR #11, opened 2025-06-04
add rotating allocator for UInt32

crosbymichael:
This creates an allocator based on a FIFO that will allocate through the range before reusing previously released allocations.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -244,6 +244,13 @@ let package = Package(
                 .product(name: "Logging", package: "swift-log"),
             ]
         ),
+        .testTarget(
+            name: "ContainerizationExtrasTests",
+            dependencies: [
+                "ContainerizationExtras",
+                "CShim",
+            ]
+        ),
         .target(
             name: "CShim"
         ),
--- a/Sources/ContainerizationExtras/NetworkAddress+Allocator.swift
+++ b/Sources/ContainerizationExtras/NetworkAddress+Allocator.swift
@@ -73,6 +73,24 @@ extension UInt32 {
             indexToAddress: { lower + UInt32($0) }
         )
     }
+
+    /// Creates a rotating allocator for vsock ports, or any UInt32 values.
+    public static func rotatingAllocator(lower: UInt32, size: UInt32) throws -> any AddressAllocator<UInt32> {
+        guard 0xffff_ffff - lower + 1 >= size else {
+            throw AllocatorError.rangeExceeded
+        }
+
+        return RotatingAddressAllocator(
+            size: size,
+            addressToIndex: { address in
+                guard address >= lower && address <= lower + UInt32(size) else {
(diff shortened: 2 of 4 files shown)


Sources/ContainerizationExtras/NetworkAddress+Allocator.swift
```diff
@@ -72,6 +72,24 @@ extension UInt32 {
             indexToAddress: { lower + UInt32($0) }
         )
     }
+
+    /// Creates an rotating allocator for vsock ports, or any UInt32 values.
```
dcantah:
```suggestion
    /// Creates a rotating allocator for vsock ports, or any UInt32 values.
```

Sources/ContainerizationExtras/RotatingAddressAllocator.swift
```diff
@@ -0,0 +1,124 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Synchronization
+
+package final class RotatingAddressAllocator: AddressAllocator {
+    package typealias AddressType = UInt32
+
+    private class State {
```
dcantah: This can be a struct from eyeballing I believe

Sources/ContainerizationExtras/RotatingAddressAllocator.swift
```diff
@@ -0,0 +1,124 @@
...
+        var allocations: [AddressType]
+        var enabled: Bool
+        var allocationCount: Int
+        let addressToIndex: AddressToIndexTransform<AddressType>
+        let indexToAddress: IndexToAddressTransform<AddressType>
+
+        init(
+            size: Int,
```
dcantah: Should size just be a UInt32 also?

Package.swift
```diff
@@ -239,6 +239,13 @@ let package = Package(
                 .product(name: "Logging", package: "swift-log"),
             ]
         ),
+        .testTarget(
```
jglogan, line 247: lol whoops

Sources/ContainerizationExtras/RotatingAddressAllocator.swift
jglogan: I'd need to switch it and see what breaks but IIRC I chose `Int` for this original because we were dealing with index values both inside the call, and outside, so it's less casting and spots where runtime failures can happen if we use Int here.

I could be wrong. It's been a while but that's what I think was the issue.

dcantah: No worries if not, just felt strange to take in an int and immediately cast it. The constructor isn't part of the protocol so we can take in whatever we want afaict

crosbymichael:
@dcantah resolved feedback

PR #21, opened 2025-06-05
Add documentation for netlink

dkovba:


Sources/ContainerizationNetlink/NetlinkSession.swift
```diff
@@ -156,17 +165,11 @@ public struct NetlinkSession {
...
-    /// ip addr add [addr] dev [interface]
-    /// ip address {add|change|replace} IFADDR dev IFNAME [ LIFETIME ] [ CONFFLAG-LIST ]
-    /// IFADDR := PREFIX | ADDR peer PREFIX
-    ///           [ broadcast ADDR ] [ anycast ADDR ]
-    ///           [ label IFNAME ] [ scope SCOPE-ID ] [ metric METRIC ]
-    /// SCOPE-ID := [ host | link | global | NUMBER ]
-    /// CONFFLAG-LIST := [ CONFFLAG-LIST ] CONFFLAG
-    /// CONFFLAG  := [ home | nodad | mngtmpaddr | noprefixroute | autojoin ]
-    /// LIFETIME := [ valid_lft LFT ] [ preferred_lft LFT ]
-    /// LFT := forever | SECONDS
```
dcantah, line 169: I wonder what'd be a good way to keep these, maybe inline in the function itself (with // comments)? These are incredibly valuable because it's nice to see what the operations in this type map to in the `ip` utility on Linux.

dcantah:
I think this just needs to rebase on `main` to get CI happy. It's just failing because of the license header updates.

Sources/ContainerizationNetlink/NetlinkSession.swift
dkovba: Done

PR #22, opened 2025-06-05
UnixSocketConfiguration: Change from/to to source/destination

dcantah:
This is something I forgot to update ages ago. The field names, while they do make sense, were different than the constructor names that eventually set the fields themselves. This is kind of awkward, and I'd much rather fix this now while it's much easier.

We'll need to update `container` after this


dcantah:
One sec, forgot to add some changes from other files

PR #30, opened 2025-06-06
Containerization: Rename nat interfaces

dcantah:
NOTE: This will be a breaking change and should be apart of a new 0.2.0 tag if we agree on this.

These were fairly generically named, and didn't give any hints to their ties to the various OS frameworks.


dcantah:
Can worry about this later

PR #33, opened 2025-06-07
LinuxProcess: Resource adjustments

dcantah:
- The kqueue handlers for stdio didn't have logic to exit on zero byte reads.
- The agent wasn't getting closed explicitly in .delete()
- In LinuxContainer we should call delete just for sanity, if for nothing more than ensuring the agent vsock fd is closed.

### DIFF
--- a/Sources/Containerization/LinuxProcess.swift
+++ b/Sources/Containerization/LinuxProcess.swift
@@ -71,15 +71,21 @@ public final class LinuxProcess: Sendable {
         var stdout: FileHandle?
         var stderr: FileHandle?
 
-        func close() throws {
+        mutating func close() throws {
             if let stdin {
                 try stdin.close()
+                stdin.readabilityHandler = nil
+                self.stdin = nil
             }
             if let stdout {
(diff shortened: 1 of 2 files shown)


Sources/Containerization/LinuxProcess.swift
```diff
@@ -208,7 +208,12 @@ extension LinuxProcess {
                 // as it always allocates. We can likely do the read loop ourselves
                 // with a buffer we allocate once on creation of the process.
                 do {
-                    try stdout.writer.write(handle.availableData)
+                    let data = handle.availableData
+                    if data.isEmpty {
+                        handle.readabilityHandler = nil
```
adityaramani: Should these be `handles[1].readabilityHandler = nil` ?

adityaramani: Or do they refer to the same thing?

dcantah: Good catch, I think we need what you'd typed.

PR #34, opened 2025-06-07
Integration: Add a multi-exec output test

dcantah:
Add a test similar to the concurrent exec test, except supply stdout.

### DIFF
--- a/Sources/Integration/ProcessTests.swift
+++ b/Sources/Integration/ProcessTests.swift
@@ -155,6 +155,71 @@ extension IntegrationSuite {
         }
     }
 
+    func testMultipleConcurrentProcessesOutput() async throws {
+        let id = "test-concurrent-processes-output"
+
+        let bs = try await bootstrap()
+        let container = LinuxContainer(
+            id,
+            rootfs: bs.rootfs,
+            vmm: bs.vmm
+        )
+        container.arguments = ["/bin/sleep", "1000"]
+
+        do {
+            try await container.create()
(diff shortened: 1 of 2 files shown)


Sources/Integration/ProcessTests.swift
```diff
@@ -155,6 +155,71 @@ extension IntegrationSuite {
...
+
+                        let status = try await exec.wait()
+                        if status != 0 {
+                            throw IntegrationError.assert(msg: "process status \(status) != 0")
+                        }
+
+                        let output = String(data: buffer.data, encoding: .utf8)
+                        guard output == "hi\(idx)\n" else {
+                            throw IntegrationError.assert(
+                                msg: "process should have returned on stdout 'hi' != '\(output!))")
```
jglogan: Is expected output `hi` or `hi\(idx)`? 

From the message it looks like expected is the former, but the test seems to check the latter?

dcantah: The latter, line 185 changes to hi\(idx)

jglogan: shouldn't the message in 202 have that then, instead of `'hi'`, on the LHS of the `!=`?

dcantah: Ahh ahh, I responded on my phone and didn't fully realize what line you were commenting on 🤣. Yes

dcantah: Fixed

PR #38, opened 2025-06-08
Set MTU to 1280 for containers.

jglogan:
- Needed for alpine containers in some instances.
- This should be configurable. Hardcoding in linkSet for now.


egernst:
1280 as advertised - thanks John.
```
eernst@2024-MBP container % container run -it alpine sh
/ # ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 1280 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: eth0: <BROADCAST,UP,LOWER_UP> mtu 1280 qdisc pfifo_fast state UP qlen 1000
    link/ether ca:52:45:53:2a:ca brd ff:ff:ff:ff:ff:ff
/ # 
```

PR #41, opened 2025-06-09
README: split prep, build and test directions into own sections

egernst:
If we split this out, it'll be easier to point developers at specific parts for setting up their build environment, versus building and testing.


egernst:
related: https://github.com/apple/container/pull/43

PR #50, opened 2025-06-09
Update README.md with minor correction

The-Alchemist:
Not to be pedantic, but "light weight" is usually considered to be one word.


dcantah (approved):
Thanks!

PR #54, opened 2025-06-09
Fix typo in README

corneliusroemer:


katiewasnothere:
@corneliusroemer Can you rebase this PR? I fixed some tests that weren't behaving the way we wanted them to :)

scyto:
i am not sure this change is right, it isn't a pre-built kernel, its compiled on make - or at least thats what the code and instructions imply?

PR #69, opened 2025-06-10
Update README.md

visualex:
typo, changed macOS 26 to Xcode 26

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ You need an Apple silicon Mac to build and run Containerization.
 To build the Containerization package, your system needs either:
 
 - macOS 15 or newer and Xcode 26 Beta
-- macOS 26 Beta 1 or newer
+- Xcode 26 Beta 1 or newer
 
 Applications built using the package will run on macOS 15 or later, but the following features are not available on macOS 15:
 


34celsius:
I don't believe this is a typo. The same requirements are listed in container's [BUILDING.md](https://github.com/apple/container/blob/main/BUILDING.md). The way this reads to me is that the software can be built on either macOS 15 or macOS 26 Beta 1.

README.md
```diff
@@ -33,7 +33,7 @@ You need an Apple silicon Mac to build and run Containerization.
 To build the Containerization package, your system needs either:
 
 - macOS 15 or newer and Xcode 26 Beta
-- macOS 26 Beta 1 or newer
+- Xcode 26 Beta 1 or newer
```
dcantah, line 36: This was intended to be macOS 26 Beta 1 so it doesn't need to be updated, but it is true that you'd need the newest Xcode as well so you could change to:
`macOS 26 Beta 1 or newer and Xcode 26 Beta or newer`

dcantah:
Going to close this out for now as what we have was what we intended, and the README has changed a bit since. Thanks for the work!

Issue #60, opened 2025-06-10
Unable to build containerization using `make all`


Omotolani98:
Building containerization binaries...
Building for debugging...
macro expansion @SendableProperty:5:9: error: 'inout sending' parameter '$0' cannot be task-isolated at end of function
`- /Users/mac/.proj/containerization/Sources/ContainerizationOS/Socket/Socket.swift:78:30: note: expanded code originates here
 76 |
 77 |     @SendableProperty
 78 |     private var _state: State
    +--- macro expansion @SendableProperty -----------------------------
    | 3 |         __state.withLock {
    | 4 |             $0!
    | 5 |         }
    |   |         |- error: 'inout sending' parameter '$0' cannot be task-isolated at end of function
    |   |         `- note: task-isolated '$0' risks causing races in between task-isolated uses and caller uses since caller assumes value is not actor isolated
    | 6 |     }
    | 7 |     set {
    +-------------------------------------------------------------------
 79 |
 80 |     public var fileDescriptor: Int32 {
[8/14] Compiling ContainerizationOS Socket.swift
make: *** [containerization] Error 1

davidfekke:
Try building in Xcode 26 beta. I was able to build using this version of Xcode.

dkovba:
@Omotolani98 Thanks for reporting! Please share output of the following commands:
```
sw_vers
/usr/bin/swift -v --version
xcrun --show-sdk-build-version
```

thisismz:
Hello,
I am encountering a build error when attempting to compile the containerization project on macOS 15.5  with Swift 6.1. The issue seems to be related to the compiler's inability to find types from the vmnet and Virtualization frameworks.
Problem:
Building the project fails with "cannot find type" errors for vmnet_network_ref and VZVmnetNetworkDeviceAttachment.
Steps to Reproduce:
Clone the apple/containerization repository.
Run make all in the root directory.
Error Output:

```
make all
Building containerization binaries...
[1/1] Planning build
Building for debugging...
/Users/thisismz/dev/containerization/Sources/Containerization/NATNetworkInterface.swift:51:24: error: cannot find type 'vmnet_network_ref' in scope
 49 |         var macAddress: String?
 50 |         #if !CURRENT_SDK
 51 |         var reference: vmnet_network_ref
    |                        `- error: cannot find type 'vmnet_network_ref' in scope
 52 |         #endif
 53 |     }

[...]

PR #91, opened 2025-06-11
Improve `@SendableProperty`

dkovba:
This PR ensures that we enter a lock inside the `@SendableProperty` implementation as soon as we access a computed property. Additionally, it mirrors the access level of the original property. Both changes are required for improved stability of progress updates in container. Additionally, it should resolve https://github.com/apple/containerization/issues/60 that occurs on certain configurations.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -22,6 +22,7 @@ import ContainerizationOCI
 import Foundation
 import Logging
 import SendableProperty
+import Synchronization
 
 import struct ContainerizationOS.Terminal
 
--- a/Sources/ContainerizationOCI/Client/RegistryClient+Fetch.swift
+++ b/Sources/ContainerizationOCI/Client/RegistryClient+Fetch.swift
@@ -177,10 +177,10 @@ extension RegistryClient {
                 while var buf = try await itr.next() {
                     let readBytes = Int64(buf.readableBytes)
                     received += readBytes
-                    await progress?([
-                        ProgressEvent(event: "add-size", value: readBytes)
-                    ])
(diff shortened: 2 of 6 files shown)


adityaramani (approved):

dcantah:
I don't fully understand why we need to ditch Synchronization's mutex, could you give more details?

dkovba:
`Mutex` has adopted `sending`:
```
func withLock<Result, E>((inout sending Value) throws(E) -> sending Result) throws(E) -> sending Result
```

This made it incompatible with computed variables:
```
macro expansion @SendableProperty:10:9: error: 'inout sending' parameter '$0' cannot be task-isolated at end of function
`- /Users/Dmitry/Apple/containerization/Sources/ContainerizationOS/Socket/Socket.swift:78:30: note: expanded code originates here
 75 |     private let _queue: DispatchQueue
 76 | 
 77 |     @SendableProperty
    |     |- note: in expansion of macro 'SendableProperty' on property '_state' here
    |     `- note: in expansion of macro 'SendableProperty' on property '_state' here
 78 |     private var _state: State
    +--- macro expansion @SendableProperty -----------------------------
    | 8 |         __state.withLock {
    | 9 |             $0 = newValue
    |10 |         }
    |   |         |- error: 'inout sending' parameter '$0' cannot be task-isolated at end of function
    |   |         `- note: task-isolated '$0' risks causing races in between task-isolated uses and caller uses since caller assumes value is not actor isolated
    |11 |     }
    |12 | }
    +-------------------------------------------------------------------
 79 | 
 80 |     public var fileDescriptor: Int32 {
[1666/1696] Compiling ContainerizationOS User.swift
make[1]: *** [all] Error 1
make: *** [vminitd] Error 2
```

PR #97, opened 2025-06-11
build: extend the swift-syntax version range

MojtabaHs:
Building Swift-Syntax can be time-consuming, but this issue has been resolved in the latest version by using a prebuilt binary. This improvement enhances CI build times and the overall quality of the package.

I’ve kept the version at 600 for compatibility reasons, but it might be possible to start from 601 as well.
Please let me know.
Thanks

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -55,7 +55,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.80.0"),
         .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.20.1"),
         .package(url: "https://github.com/apple/swift-system.git", from: "1.4.0"),
-        .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "600.0.0-latest"),
+        .package(url: "https://github.com/swiftlang/swift-syntax.git", "600.0.0"..<"602.0.0"),
         .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.1.0"),
     ],
     targets: [
--- a/Package.swift
+++ b/Package.swift
@@ -55,7 +55,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.80.0"),
         .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.20.1"),
         .package(url: "https://github.com/apple/swift-system.git", from: "1.4.0"),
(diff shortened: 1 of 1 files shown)


crosbymichael:
@dkovba Can you review this change today?

Package.swift
```diff
@@ -55,7 +55,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.80.0"),
         .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.20.1"),
         .package(url: "https://github.com/apple/swift-system.git", from: "1.4.0"),
-        .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "600.0.0-latest"),
+        .package(url: "https://github.com/swiftlang/swift-syntax.git", "600.0.0"..<"602.0.0"),
```
dkovba: Please consider changing to `from: "600.0.0"`. This should be sufficient to ensure that a package prebuilt is downloaded.

MojtabaHs: I’ve updated it as you guided.

Since `601` has already been released and `602` is just around the corner in `prerelease`, we might want to update the dependency to support the highest possible version, while still maintaining compatibility with older versions if needed.

Let me know if I can assist further.
Thanks!

dkovba: Thank you for the explanation! We can get a package prebuilt with `from: "600.0.0"`. But since we're declaring `// swift-tools-version: 6.0` in `Package.swift` and use Swift 6.1, the originally suggested change to `"600.0.0"..<"602.0.0"` makes a perfect sense, and I'll be happy to approve it.

dkovba: Since Xcode 26 beta is required for building Containerization, we may consider `"600.0.0"..<"603.0.0"` as well. This is the version of Swift that comes with Xcode 26 beta:
```
% /usr/bin/swift -v --version
Apple Swift version 6.2 (swiftlang-6.2.0.9.909 clang-1700.3.9.907)
Target: arm64-apple-macosx15.0
swift-driver version: 1.127.4.2 /Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-frontend --version
Apple Swift version 6.2 (swiftlang-6.2.0.9.909 clang-1700.3.9.907)
Target: arm64-apple-macosx15.0
```

MojtabaHs: Extended! 
If we can support 3 major versions, then why not? 💪🏻

dkovba (approved):

PR #106, opened 2025-06-12
gitignore: Remove editor specific git ignore rules

dcantah:
These truthfully should be up to users to maintain, as trying to add everyones editor of choice config directories here isn't ideal.

This additionally adds a section in contributing on the new stance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,9 +10,6 @@ DerivedData/
 .swiftpm
 workdir/
 installer/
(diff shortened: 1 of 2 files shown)


dcantah:
@KeoFoxy Sorry about this! We thought about this some more and if we supported everyones editor of choice the list can get quite long, we believe this is better suited to be handled by the developers themselves that want to contribute.

PR #108, opened 2025-06-12
fix: inconsistent capitalisation in README

rua-iri:
The capitalisation in the README file for the section headings was not consistent across all sections.

This PR ensures that each word in the headings is consistently capitalised, with the exception of prepositions.


jglogan:
We don't have a style guide published for our documentation, but our headings use sentence case, not title case.

jglogan:
@rua-iri I'm closing this PR, but do you see any inconsistencies where there are title case headers that should be in sentence case? We'd be happy to take corrections for those!

PR #110, opened 2025-06-12
Add a way to detect if a process' io has completed

adityaramani:
This change adds a new private method  `waitIoComplete` on the `LinuxProcess` type.

This method is called internally when the user calls `wait` for a process - and it tries to give the IO streams some time to clear their buffers.

Internally, this method sets up an `AsyncStream` down which an item is sent when the vsock connection for either stdout/stderr is terminated.
We get this termination signal when the readability handler for the associated fd fires with a no available data.

Inside the guest - once we are done relaying the IO from the process into the socket connection, we close the socket fd which triggers the above.

All this logic is wrapped around a timeout of 3 seconds, just to ensure the method does not block forever.

### DIFF
--- a/Sources/Containerization/LinuxProcess.swift
+++ b/Sources/Containerization/LinuxProcess.swift
@@ -99,6 +99,7 @@ public final class LinuxProcess: Sendable {
         struct IoTracker {
             let stream: AsyncStream<Void>
             let cont: AsyncStream<Void>.Continuation
+            let configuredStreams: Int
         }
     }
 
@@ -213,18 +214,16 @@ extension LinuxProcess {
             }
         }
 
-        var setOutputRelay = false
+        var configuredStreams = 0
         let (stream, cc) = AsyncStream<Void>.makeStream()
         if let stdout = self.ioSetup.stdout {
-            setOutputRelay = true
+            configuredStreams += 1
             handles[1]?.readabilityHandler = { handle in
-                // NOTE: We need some way to know when this data is done being piped,
-                // so DispatchGroup or similar. `availableData` is also pretty poor,
-                // as it always allocates. We can likely do the read loop ourselves
-                // with a buffer we allocate once on creation of the process.
                 do {
                     let data = handle.availableData
                     if data.isEmpty {
+                        // This block is called when the producer (the guest) closes
+                        // the fd it is writing into.
                         handles[1]?.readabilityHandler = nil
                         cc.yield()
                         return
@@ -237,7 +236,7 @@ extension LinuxProcess {
         }
 
         if let stderr = self.ioSetup.stderr {
-            setOutputRelay = true
+            configuredStreams += 1
             handles[2]?.readabilityHandler = { handle in
                 do {
                     let data = handle.availableData
@@ -252,9 +251,9 @@ extension LinuxProcess {
                 }
             }
         }
-        if setOutputRelay {
+        if configuredStreams > 0 {
             self.state.withLock {
-                $0.ioTracker = .init(stream: stream, cont: cc)
+                $0.ioTracker = .init(stream: stream, cont: cc, configuredStreams: configuredStreams)
             }
         }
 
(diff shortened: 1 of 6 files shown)


Sources/Containerization/LinuxProcess.swift
```diff
@@ -318,11 +335,13 @@ extension LinuxProcess {
     @discardableResult
     public func wait(timeoutInSeconds: Int64? = nil) async throws -> Int32 {
         do {
-            return try await self.agent.waitProcess(
+            let code = try await self.agent.waitProcess(
                 id: self.id,
                 containerID: self.owningContainer,
                 timeoutInSeconds: timeoutInSeconds
             )
+            try await self.waitIoComplete()
```
adityaramani: Should we add this here? Or make the method public - allowing for users to optionally call it?

Sources/Containerization/LinuxProcess.swift
```diff
@@ -335,16 +354,50 @@ extension LinuxProcess {
         }
     }
 
+    /// Wait until the standard output and standard error streams for the process have concluded.
+    private func waitIoComplete() async throws {
+        let ioTracker = self.state.withLock { $0.ioTracker }
+        guard let ioTracker else {
+            return
+        }
+        let required = {
```
dcantah: We might be able to do this dance before creating the object and just pass directly to the object itself how many streams we need to wait for. The guard below (although honestly maybe being defensive here is good, but it seems a bit much) seems unlikely to ever be hit, as we only create the tracker if we asked for stdout/stderr

Sources/Containerization/LinuxProcess.swift
```diff
@@ -335,16 +354,50 @@ extension LinuxProcess {
...
+            if self.ioSetup.stdout != nil {
+                num += 1
+            }
+            return num
+        }()
+        guard required > 0 else {
+            return
+        }
+        do {
+            try await Timeout.run(seconds: 3) {
```
dcantah, line 363: I forgot we had this 😆

vminitd/Sources/vminitd/TerminalIO.swift
```diff
@@ -139,15 +138,16 @@ final class TerminalIO: ManagedProcess.IO & Sendable {
...
             // We could alternatively just allocate buffers in the constructor, and free them
             // on close().
             buffer.deallocate()
-            try ProcessSupervisor.default.poller.delete(fd)
+            try ProcessSupervisor.default.poller.delete(readFd)
         } catch {
-            self.log?.error("failed to delete pipe fd from epoll \(fd): \(error)")
+            self.log?.error("failed to delete pipe fd from epoll \(readFd): \(error)")
         }
+        Foundation.close(writeFd)
```
dcantah: nit: Check if < 0 and log the errno just in case. These io issues can get gnarly to look into

dcantah: Same for StandardIO.swift

vminitd/Sources/vminitd/ProcessSupervisor.swift
```diff
@@ -98,7 +98,7 @@ actor ProcessSupervisor {
 
             return try process.start()
         } catch {
-            self.log?.error("process start failed \(error)")
+            self.log?.error("process \(process.id) start failed \(error)")
```
dcantah: nit: Can you put the process ID as metadata?

Sources/Containerization/LinuxProcess.swift
dcantah: I think this is fine, I'd like to avoid more things the user needs to worry about.

dcantah (commented):
Looks great! Couple comments

Sources/Containerization/LinuxProcess.swift
```diff
@@ -335,16 +353,37 @@ extension LinuxProcess {
         }
     }
 
+    /// Wait until the standard output and standard error streams for the process have concluded.
+    private func waitIoComplete() async throws {
```
dcantah: Does this need to be throwing? It seems the only way for it to fail is the CancellationError, but all we do is log

dcantah (approved):
One comment, LGTM!

dcantah (approved):

Sources/Containerization/LinuxProcess.swift
```diff
@@ -335,16 +353,37 @@ extension LinuxProcess {
...
     }
 
+    /// Wait until the standard output and standard error streams for the process have concluded.
+    private func waitIoComplete() async {
+        let ioTracker = self.state.withLock { $0.ioTracker }
+        guard let ioTracker else {
+            return
+        }
+        do {
+            try await Timeout.run(seconds: 3) {
```
crosbymichael, line 363: What does this mean in practice?

crosbymichael (approved):

PR #111, opened 2025-06-13
LinuxProcess: Always return CZError

dcantah:
We should return our own error type always for our methods. This wraps all of the public methods on the process to return a more full featured error than the rpcerrors we get today.

### DIFF
--- a/Sources/Containerization/LinuxProcess.swift
+++ b/Sources/Containerization/LinuxProcess.swift
@@ -262,72 +262,98 @@ extension LinuxProcess {
 
     /// Start the process.
     public func start() async throws {
-        let spec = self.state.withLock { $0.spec }
-
-        var streams = [VsockConnectionStream?](repeating: nil, count: 3)
-        if let stdin = self.ioSetup.stdin {
-            streams[0] = try self.vm.listen(stdin.port)
-        }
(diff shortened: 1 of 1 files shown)


Sources/Containerization/LinuxProcess.swift
```diff
@@ -246,72 +246,95 @@ extension LinuxProcess {
...
-                stderr: result[2]
+            self.state.withLock {
+                $0.stdio = StdioHandles(
+                    stdin: result[0],
+                    stdout: result[1],
+                    stderr: result[2]
+                )
+                $0.pid = pid
+            }
+        } catch {
```
crosbymichael, line 312: Since some of the methods we are calling in the `do {` also throw CZerror with a more exact code and message, should we improve this `catch` block to also catch `ContainerizationError` and just rethrow in that case?

crosbymichael (changes requested):

Sources/Containerization/LinuxProcess.swift
dcantah: Agree, I did this for other methods before like `wait`, I didn't realize things in this block returned czerror already though. Fixin!

crosbymichael (approved):

PR #112, opened 2025-06-13
EXT4: Adjustments to file names and visibilities

dcantah:
- FileTimestamps constructor that actually let you provide values was not public, so it wasn't possible to pass in values other than nil. This change makes the other constructor and the underlying fields public.
- Rename EXT4+Format to EXT4+Formatter
- Rename EXT4+Export to EXT4Reader+Export
- Make the superblock publicly accessible in the reader like the docs for the product states.


dcantah:
Will fix formatting shortly

PR #126, opened 2025-06-14
LinuxContainer: Add run method

dcantah:
To cut down on some of the boilerplate for running containers we can have a run method which is common on a lot of process wrappers. This method does `create + start + wait` for you, and optionally cleans up (stops) the container as well if you provide true for the only argument named `stop`.


dcantah:
@crosbymichael I vaguely remember pitching this before, and I don't remember if you hated it or enjoyed it, but here it is in textual format for you to review 😄. The main argument would be users could make this with an extension, but I'd argue it's the simplest way to run a quick container using this library and worth just being a method without a user crafting it themselves.

Sources/Containerization/LinuxContainer.swift
```diff
@@ -552,6 +552,19 @@ extension LinuxContainer {
         }
     }
 
+    /// Run runs the container until the init process exits, and optionally
+    /// tears down the container after. This is equivalent to:
+    /// create() + start() + wait() (+ stop() if `stop` is set to true).
+    public func run(stop: Bool = false) async throws -> Int32 {
```
crosbymichael: What if we remove the `stop` param or it defaults to `true`?  I believe it is ok to expect a function that calls create/start to also be responsible for calling stop (default wise).

dcantah: Works for me!

dcantah: Removed the stop param and just always do stop now

Sources/Containerization/LinuxContainer.swift
```diff
@@ -552,6 +552,18 @@ extension LinuxContainer {
...
     }
 
+    /// Run runs the container until the init process exits, and
+    /// tears down the container after. This is equivalent to:
+    /// create() + start() + wait() + stop().
+    public func run(stop: Bool = true) async throws -> Int32 {
+        try await self.create()
+        try await self.start()
+        let status = try await self.wait()
+        try await self.stop()
```
adityaramani, line 562: Need to check for `if stop` ? Or remove the param?

dcantah: I guess my local changes got lost? Will push when back by pc

dcantah: Fixed!

adityaramani (approved):

crosbymichael:
Let's close this one now as we explore a more structured API for a `run()` method.

PR #129, opened 2025-06-14
Adds some context to fetchToken errors

makhov:
Fixes https://github.com/apple/container/issues/182 and https://github.com/apple/container/issues/183

### DIFF
--- a/Sources/ContainerizationOCI/Client/RegistryClient.swift
+++ b/Sources/ContainerizationOCI/Client/RegistryClient.swift
@@ -184,13 +184,26 @@ public final class RegistryClient: ContentClient {
                     if let ct = currentToken, ct.isValid(scope: tokenRequest.scope) {
                         break
                     }
-                    let _currentToken = try await fetchToken(request: tokenRequest)
-                    guard let token = _currentToken.getToken() else {
-                        throw ContainerizationError(.internalError, message: "Failed to fetch Bearer token")
+
+                    do {
+                        let _currentToken = try await fetchToken(request: tokenRequest)
+                        guard let token = _currentToken.getToken() else {
+                            throw ContainerizationError(.internalError, message: "Failed to fetch Bearer token")
+                        }
+                        currentToken = _currentToken
+                        request.headers.replaceOrAdd(name: "Authorization", value: token)
+                        retryCount += 1
+                    } catch let err as RegistryClient.Error {
+                        guard case .invalidStatus(_, let status, _) = err else {
+                            throw err
+                        }
+                        if status == .unauthorized || status == .forbidden {
+                            throw RegistryClient.Error.invalidStatus(url: path, _response.status, reason: "Access denied or wrong credentials")
+                        }
+
+                        throw err
                     }
-                    currentToken = _currentToken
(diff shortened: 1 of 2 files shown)


Tests/ContainerizationOCITests/RegistryClientTests.swift
```diff
@@ -80,11 +80,9 @@ struct OCIClientTests: ~Copyable {
...
         let authentication = BasicAuthentication(username: "foo", password: "bar")
         let client = RegistryClient(host: "ghcr.io", authentication: authentication)
-        let error = await #expect(throws: RegistryClient.Error.self) { try await client.ping() }
-        if case .invalidStatus = error {
-        } else {
-            Issue.record("encountered unexpected error \(error)")
-        }
+        let error = await #expect(throws: ContainerizationError.self) { try await client.ping() }
+        #expect(error?.code == .internalError)
+        #expect(error?.message == "HTTP request to https://ghcr.io/token?client_id=containerization-registry-client&service=ghcr.io&scope=repository:user/image:pull failed with response: 403 Forbidden. Access denied or wrong credentials.")
```
makhov: Not the best way to do it, but I couldn't come up with something better

Sources/ContainerizationOCI/Client/RegistryClient.swift
```diff
@@ -184,13 +184,26 @@ public final class RegistryClient: ContentClient {
...
+                        }
+                        currentToken = _currentToken
+                        request.headers.replaceOrAdd(name: "Authorization", value: token)
+                        retryCount += 1
+                    } catch let err as RegistryClient.Error {
+                        guard case .invalidStatus(_, let status) = err else {
+                            throw err
+                        }
+                        if status == .unauthorized || status == .forbidden {
+                            throw ContainerizationError(.internalError, message: "\(String(describing: err)). Access denied or wrong credentials.")
```
adityaramani: Should we re-throw `RegistryClient.Error.invalidStatus` here? But just add the wording you have?

Then in your test below we can check for the `invalidStatus` error?

makhov: Sound good, but `RegistryClient.Error.invalidStatus` doesn't any `message` field, so I'm not sure how to add the context. Should I add an optional field, that will be used in `description`?

adityaramani: Ah sorry yeah. I had a change locally where I was going to put in a additional message field to that error. Let me pr that real quick.

Cause I think the ImageService back in container checks for the status code as well

adityaramani: Want to take a look at https://github.com/apple/containerization/pull/134? Do you think this would be a good addition?

makhov: I've updated the PR to support your changes

adityaramani (approved):

dcantah (approved):
Thanks!

dcantah:
@makhov Can you sign the first commit and re-push? We can get this in right after

makhov:
Thanks! I squashed them and signed-off the commit.

PR #134, opened 2025-06-15
More informative errors from RegistryClient

adityaramani:
Also creates an `ErrorResponse` type to model the errors typically returned by a container registry. 

Reference: https://distribution.github.io/distribution/spec/api/#errors

Example error message:
```
Error: HTTP request to https://ghcr.io/token?client_id=containerization-registry-client&service=ghcr.io&scope=repository:user/image:pull failed with response: 403 Forbidden. Reason: {"errors":[{"message":"requested access to the resource is denied","code":"DENIED"}]}
```


makhov (commented):
LGTM

PR #140, opened 2025-06-16
Fix `make release` goal

dcantah:
We were using a := assignment for BUILD_BIN_DIR which evaluates any variables immediately. Our make release logic sets the envvar and then invokes `all` but at that point the value of BUILD_BIN_DIR has already been determined.


dkovba (approved):
👍

PR #143, opened 2025-06-17
implement stats command

### LINKED ISSUE #138: [Request]: Container Stats Functionality

Docker has the ability to show usage statistics for container utilization (cpu %, memory usage, network i/o, block i/o, etc). 

I'd love to see this make it's way to the containerization framework, and am curious how the implementation might work. I know Docker on Linux uses cgroups for this, which doesn't exist in Darwin. I might assume it would require more work in Virtualization.framework to expose that information better (other than simply watching the `Virtual Machine Service for X` process). 

Either way would love to have it, would be interested to help provide this functionality if Apple unblocks a path forward!

### Code of Conduct

elijah-wright:
this PR implements a stats command by reading from the `cgroup` path

closes #138


elijah-wright:
@dcantah I ran `make fmt` and the linting tests should work now

dcantah:
@elijah-wright Thanks for this! We appreciate it :). I should've been more clear on the issue from the start that this functionality we'd ideally like to go through some design review for rather than jumping right in though. I think there is some merit to what the author of the issue posted in that it's possible we could find a way to more accurately convey what the total resource usage is, as there is a bit more than just the workload in the guest that is really happening (there's more JUST in the guest itself disregarding the workload, as vminitd/the guest agent uses some resources as well).

elijah-wright:
:+1: yeah I think the base is there in this PR but I don't know about the final output, `docker stats` has a nice design

I was looking at `Linux` and I saw `cgroupsPath`, which I set to the container ID in `createDefaultRuntimeSpec`. do correct me if I'm wrong but I think that would fix the `vminitd` issue, since the agent PID should rest outside of the path

elijah-wright:
I'm going to close this because I accidentally did this on the main branch and because I think more discussion should happen first

PR #149, opened 2025-06-17
Sort keys in image index to keep the digest consistent

makhov:
The digest is calculated based on written json data, but JSONEncoder doesn't preserve order bby default, so the digest is not consistent.


makhov:
Done!

PR #151, opened 2025-06-17
Create unpacker protocol + ext4 unpacker

adityaramani:
Creates a new Unpacker protocol that defines a single method

```
func unpack(_ image: Image, for platform: Platform, at path: URL, progress: ProgressHandler?) async throws -> Mount
```

This change also removes the `unpack(...)` method from the Image type.

Before
```
let mount = try await image.unpack(for: platform, at: path)
```
After
```
let unpacker = EXT4Unpacker(blockSizeInBytes: 2.gib())
let mount = try await unpacker.unpack(image, for: platform, at: path)
```

### DIFF
--- a/Sources/Containerization/Image/Unpacker/EXT4Unpacker.swift
+++ b/Sources/Containerization/Image/Unpacker/EXT4Unpacker.swift
@@ -45,24 +45,21 @@ public struct EXT4Unpacker: Unpacker {
             try Task.checkCancellation()
             let content = try await image.getContent(digest: layer.digest)
 
+            let compression: ContainerizationArchive.Filter
             switch layer.mediaType {
             case MediaTypes.imageLayer, MediaTypes.dockerImageLayer:
-                try filesystem.unpack(
-                    source: content.path,
-                    format: .paxRestricted,
-                    compression: .none,
-                    progress: progress
-                )
+                compression = .none
             case MediaTypes.imageLayerGzip, MediaTypes.dockerImageLayerGzip:
-                try filesystem.unpack(
-                    source: content.path,
-                    format: .paxRestricted,
-                    compression: .gzip,
-                    progress: progress
-                )
+                compression = .gzip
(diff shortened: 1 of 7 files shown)


Sources/Containerization/Image/Unpacker/EXT4Unpacker.swift
```diff
@@ -0,0 +1,84 @@
...
+                try filesystem.unpack(
+                    source: content.path,
+                    format: .paxRestricted,
+                    compression: .gzip,
+                    progress: progress
+                )
+            default:
+                throw ContainerizationError(.unsupported, message: "Media type \(layer.mediaType) not supported.")
+            }
+        }
```
dcantah: Nit: Given both the unpacks are the exact same besides whether they supply compression or not, can we just assign a compression variable in the switch cases and then do the unpack outside of it?

Sources/cctl/ImageCommand.swift
```diff
@@ -126,11 +133,20 @@ extension Application {
                 }
 
                 print("image pulled")
+                guard let unpackPath else {
+                    return
+                }
+                guard !FileManager.default.fileExists(atPath: unpackPath) else {
```
dcantah, line 139: I think this also returns true if the path is a directory, do we still want to error?

dcantah (approved):
Looks great :)

Sources/Containerization/Image/Unpacker/EXT4Unpacker.swift
adityaramani: Yep. Updated

Sources/cctl/ImageCommand.swift
adityaramani: I changed the wording to say that `--unpack-path` should be a new directory and updated the error to match

crosbymichael (approved):

PR #153, opened 2025-06-18
run cleanupRelay only once in TerminalIO and StandardIO

### LINKED ISSUE #124: [Bug]: The CI has a high failure rate
### I have done the following

### Steps to reproduce

1. Trigger PR checks.

### Current behavior

There is a high chance of getting this:
```
2025-06-13T23:58:46+0000 error com.apple.containerization : [containerization_integration] ❌ test multiple concurrent processes with output stress failed: internal error (13): startProcess: failed to start process: Error Domain=NSPOSIXErrorDomain Code=9 "Bad file descriptor"
2025-06-13T23:58:46+0000 info com.apple.containerization : [containerization_integration] 
Integration suite completed in 19.997790932655334s with 8/10 passed and 1/10 skipped!
2025-06-13T23:58:46+0000 error com.apple.containerization : [containerization_integration] ❌
```

### Expected Behavior

The tests pass.

### Environment

```markdown
Reproducible in CI only.
```

### Code of Conduct

elijah-wright:
possibly fixes #124

### DIFF
--- a/vminitd/Sources/vminitd/ProcessSupervisor.swift
+++ b/vminitd/Sources/vminitd/ProcessSupervisor.swift
@@ -64,10 +64,15 @@ actor ProcessSupervisor {
         let exited = Reaper.reap()
         self.log?.debug("finished wait4 of \(exited.count) processes")
 
-        for proc in processes {
-            let pid = proc.pid
-            self.log?.debug("checking for exit of managed process", metadata: ["pid": "\(pid)", "exits": "\(exited)"])
+        self.log?.debug("checking for exit of managed process", metadata: ["exits": "\(exited)", "processes": "\(processes.count)"])
+        let exitedProcesses = self.processes.filter { proc in
+            exited.contains { pid, _ in
+                proc.pid == pid
+            }
+        }
 
+        for proc in exitedProcesses {
+            let pid = proc.pid
             if pid <= 0 {
                 continue
             }
@@ -80,7 +85,6 @@ actor ProcessSupervisor {
                         "status": "\(status)",
                         "count": "\(processes.count - 1)",
                     ])
-
                 proc.setExit(status)
                 self.processes.removeAll(where: { $0.pid == pid })
             }
@@ -95,7 +99,6 @@ actor ProcessSupervisor {
 
         do {
             self.processes.append(process)
-
             return try process.start()
         } catch {
             self.log?.error("process start failed \(error)", metadata: ["process-id": "\(process.id)"])
--- a/vminitd/Sources/vminitd/StandardIO.swift
+++ b/vminitd/Sources/vminitd/StandardIO.swift
@@ -119,6 +119,10 @@ final class StandardIO: ManagedProcess.IO & Sendable {
         }
     }
 
+    final class RelayState: @unchecked Sendable {
+        var done = false
+    }
+
     // NOP
     func resize(size: Terminal.Size) throws {}
 
@@ -128,10 +132,10 @@ final class StandardIO: ManagedProcess.IO & Sendable {
         // `buf` isn't used concurrently.
         nonisolated(unsafe) let buf = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: Int(getpagesize()))
 
-        var didCleanup = false
-        let cleanupRelay: @Sendable () -> Void = {
-            if didCleanup { return }
-            didCleanup = true
+        let state = RelayState()
+        let cleanupRelay: @Sendable () -> Void = { [state] in 
+            guard !state.done else { return }
+            state.done = true
             self.cleanupRelay(readFd: readFromFd, writeFd: writeToFd, buffer: buf, log: self.log)
         }
 
--- a/vminitd/Sources/vminitd/StandardIO.swift
+++ b/vminitd/Sources/vminitd/StandardIO.swift
@@ -78,7 +78,7 @@ final class StandardIO: ManagedProcess.IO & Sendable {
                 port: stdinPort,
                 cid: VsockType.hostCID
             )
-            let stdinSocket = try Socket(type: type)
+            let stdinSocket = try Socket(type: type, closeOnDeinit: false)
             try stdinSocket.connect()
             self.stdinSocket = stdinSocket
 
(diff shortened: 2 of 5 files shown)


elijah-wright:
@adityaramani I think the formatting issues have been fixed

adityaramani:
Im gonna pull in the changes locally and test it out as well

elijah-wright:
thank you, I've been depending on the workflow tests and it's not fair to you that they're not passing

adityaramani:
No worries at all!

adityaramani:
Tried running the tests in a loop and it failed with this, and there werent any other logs from the guest
```
2025-06-18T14:50:23-0700 error com.apple.containerization : [containerization_integration] ❌ test multiple concurrent processes with output stress failed: internalError: "failed to start process" (cause: "internal error (13): startProcess: failed to start process: Error Domain=NSCocoaErrorDomain Code=256 "(null)"UserInfo={NSUnderlyingError=Error Domain=NSPOSIXErrorDomain Code=9 "Bad file descriptor"}")
```

I'll try combining your change with what I was testing out with the `writeFd` being deinited

vminitd/Sources/vminitd/StandardIO.swift
```diff
@@ -119,6 +119,10 @@ final class StandardIO: ManagedProcess.IO & Sendable {
         }
     }
 
+    final class RelayState: @unchecked Sendable {
```
dcantah: I wonder, can this be a struct?

dcantah: Well truthfully, although the Swift gods will scorn us, we know with the current implementation that the epoll callback should never run concurrently, so we could likely use an inline bool in StandardIO and TerminalIO. You'll probably have to mark it as nonisolated(unsafe) however

dcantah: The same scheme we use for the buffers for both the relays

elijah-wright:
dang, I really thought that was going to solve it. I'm going to look at this again when I get home, I got caught in a literal sandpit 💀

vminitd/Sources/vminitd/StandardIO.swift
adityaramani: I think we can use the `@SendableProperty` macro we have and define a boolean
```
    @SendableProperty
    private var closed: Bool = false
...

    func cleanupRelay(...) {
        guard !self.closed else {
            return
        }
        ...
```

dcantah: That also works, I just don't think we actually need to lock here today. But realistically this would be completely negligible 🐸, and if we ever changed the behavior of the epoll callbacks `SendableProperty` is probably the better option, but I'd hope if the behavior changed we'd make that configurable in the api itself

elijah-wright: > I wonder, can this be a struct? 

I created this as a class with a function before I realized that it didn't need to be a function but I didn't change its type

and yeah I did look at `NSLock` but I figured it was ok not to use it since this isn't multithreaded

dcantah: >and yeah I did look at NSLock but I figured it was ok not to use it since this isn't multithreaded

I also think it's fine given the epoll implementation we have. Wanna trial the inline bool idea?

elijah-wright: ok I did 👍

vminitd/Sources/vminitd/StandardIO.swift
```diff
@@ -128,9 +128,16 @@ final class StandardIO: ManagedProcess.IO & Sendable {
         // `buf` isn't used concurrently.
         nonisolated(unsafe) let buf = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: Int(getpagesize()))
 
+        nonisolated(unsafe) var didCleanup = false
```
dkovba, line 131: Is `didCleanup` thread-safe? Should we access it within a lock? The same applies to `TerminalIO.swift`.

elijah-wright: I think @dcantah and I talked about this, it should be fine to use an inline bool here since there's no multithreading

adityaramani:
Thanks! I think your changes look good - and I tested it after adding in a couple of other fixes (see here https://github.com/adityaramani/containerization/tree/vminitd-crash) 

Although I do think there maybe some other issues in the process handling flow,  this change makes it more stable.

Will wait for the others to review

vminitd/Sources/vminitd/StandardIO.swift
dkovba: It might be worth adding `didCleanup` into the comment about thread-safety of `buf`.

elijah-wright: I did 👍

elijah-wright:
@adityaramani do you want me to cherry pick your commit? it's only partially signed so I don't know

adityaramani:
Yeah go for it. Take that commit in. If the signature are a problem just apply the diff

elijah-wright:
ok done, but it says unverified

adityaramani:
Yeah. Maybe try removing me as author?

elijah-wright:
@adityaramani if I could ask, what are the specs of the machine you're testing on?

adityaramani:
It's an M2 Pro with 64 gb of ram. Are you asking cause you don't see the issue on your end?

I need to run the test on a loop and see it failing once in like 15 times

elijah-wright:
I was just wondering, I think it has to be a race condition then. increasing system load (`stress-ng`) might make it more predictable for testing if that's true

vminitd/Sources/vminitd/ProcessSupervisor.swift
```diff
@@ -64,10 +64,15 @@ actor ProcessSupervisor {
         let exited = Reaper.reap()
         self.log?.debug("finished wait4 of \(exited.count) processes")
 
-        for proc in processes {
-            let pid = proc.pid
-            self.log?.debug("checking for exit of managed process", metadata: ["pid": "\(pid)", "exits": "\(exited)"])
+        self.log?.debug("checking for exit of managed process", metadata: ["exits": "\(exited)", "processes": "\(processes.count)"])
```
dcantah, line 67: Great catch 😆

dcantah (approved):
This looks good to me, but I'll wait for either @adityaramani or @crosbymichael to give it a second thumbs up.

adityaramani:
LGTM! I think Michael should take a look as well

PR #154, opened 2025-06-19
Use case-insensitive check for www-authenticate header

makhov:
Probably fixes https://github.com/apple/container/issues/237

Anyway, the check should be case-insensitive

### DIFF
--- a/Sources/ContainerizationOCI/Client/RegistryClient.swift
+++ b/Sources/ContainerizationOCI/Client/RegistryClient.swift
@@ -171,10 +171,12 @@ public final class RegistryClient: ContentClient {
                 let _response = try await client.execute(request, deadline: .distantFuture)
                 response = _response
                 if _response.status == .unauthorized || _response.status == .forbidden {
-                    let authHeader = _response.headers[TokenRequest.authenticateHeaderName]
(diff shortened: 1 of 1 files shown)


Sources/ContainerizationOCI/Client/RegistryClient.swift
```diff
@@ -171,10 +171,12 @@ public final class RegistryClient: ContentClient {
                 let _response = try await client.execute(request, deadline: .distantFuture)
                 response = _response
                 if _response.status == .unauthorized || _response.status == .forbidden {
-                    let authHeader = _response.headers[TokenRequest.authenticateHeaderName]
+                    guard let authHeader = _response.headers.first(name: TokenRequest.authenticateHeaderName) else {
```
adityaramani, line 174: I think the existing subscript method is also a case-insensitive check?

This is what it ends up as
https://github.com/apple/swift-nio/blob/main/Sources/NIOHTTP1/HTTPTypes.swift#L447

makhov: Right, my bad, completely missed it, sorry. I close the PR then.

adityaramani: No worries!

PR #156, opened 2025-06-20
Configure only a single gateway if there are multiple interfaces.

jglogan:


Sources/Containerization/LinuxContainer.swift
```diff
@@ -500,12 +500,14 @@ extension LinuxContainer {
...
-                // 3. Add the gateway address
                 for (index, i) in self.interfaces.enumerated() {
                     let name = "eth\(index)"
                     try await agent.addressAdd(name: name, address: i.address)
                     try await agent.up(name: name)
-                    try await agent.routeAddDefault(name: name, gateway: i.gateway)
+                }
+                // Configure the gateway only on the first interface
+                if !self.interfaces.isEmpty {
+                    try await agent.routeAddDefault(name: "eth0", gateway: self.interfaces[0].gateway)
```
katiewasnothere: I feel like we should just be configuring whatever is given to us here. Like if the network has a gateway on it, we configure it. There could be cases where we want all the gateways configured as routes and only a single one as the default gateway.

jglogan: That makes sense, thanks! IMO the default policy should prevent folks from shooting themselves in the foot with multiple default routes, but it makes sense that that policy should exist in the sandbox service and not here. I'll look into how to rework the PR to do that without blowing up the scope of this change.

crosbymichael: I agree with making the `gateway` optional and only configuring it when provided.  That still is not a great design because of the "default" route.  I don't know if we should add any additional information to the interface type that would denote the "default" route.

jglogan: Agree, let's do the minimum in this one and get the mechanism/policy split correct. We can consider how Interface evolves over time elsewhere.

PR #166, opened 2025-06-24
Makefile: Make target deps more explicit

### LINKED ISSUE #165: [Request]: Make target dependencies should be made explicit

Some of the `make` targets depend on other targets, but this is not visible anywhere. We should make the dependencies explicit.

`init` should depend on `containerization` because it needs `bin/cctl` installed
`integration` depends on `init` otherwise tests fail with "Invalid domain for image reference vminit:latest"

### Code of Conduct

dcantah:
Fixes #165

Some of the targets rely on others but weren't actually listed:
- `init` relies on `containerization` as it invokes `cctl`.
- `integration` relies on `init` as it needs the rootfs with vminitd inside.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ containerization:
 	@codesign --force --sign - --timestamp=none --entitlements=signing/vz.entitlements bin/containerization-integration
 
 .PHONY: init
-init: vminitd
+init: containerization vminitd
 	@echo Creating init.ext4...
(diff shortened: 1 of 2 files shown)


madrob:
I think we need another layer of indirection, you don't want to depend on `.PHONY` targets as those will always be considered out of date. We should have intermediate rules that specify the files we _actually_ need. I can work on this and show you what I mean.

dcantah:
@madrob ack go for it

Issue #119, opened 2025-06-13
[Bug]: doesn't compile with Apple Swift version 6.2-dev


cluster2600:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

While attempting to compile on macOS 26

### Current behavior

/Users/maxime/Documents/containerization/.build/checkouts/swift-nio/Package.swift:20:14: error: failed to build module 'Foundation'; this SDK is not supported by the compiler (the SDK is built with 'Apple Swift version 6.2 effective-5.10 (swiftlang-6.2.0.9.904 clang-1700.3.9.904)', while this compiler is 'Apple Swift version 6.2-dev effective-5.10 (LLVM 3c4a54b02dd62b7, Swift a40a7be6945a895)'). Please select a toolchain which matches the SDK.
 18 | // Used only for environment variables, does not make its way
 19 | // into the product code.
 20 | import class Foundation.ProcessInfo
    |              `- error: failed to build module 'Foundation'; this SDK is not supported by the compiler (the SDK is built with 'Apple Swift version 6.2 effective-5.10 (swiftlang-6.2.0.9.904 clang-1700.3.9.904)', while this compiler is 'Apple Swift version 6.2-dev effective-5.10 (LLVM 3c4a54b02dd62b7, Swift a40a7be6945a895)'). Please select a toolchain which matches the SDK.
 21 | 
 22 | let swiftAtomics: PackageDescription.Target.Dependency = .product(name: "Atomics", package: "swift-atomics")

/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Foundation.framework/Modules/Foundation.swiftmodule/arm64e-apple-macos.swiftinterface: 0:
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Foundation.framework/Modules/Foundation.swiftmodule/arm64e-apple-macos.swiftinterface: 0:
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Foundation.framework/Modules/Foundation.swiftmodule/arm64e-apple-macos.swiftinterface: 0:
error: ExitCode(rawValue: 1)

### Expected Behavior

Compilation should succeed

### Environment

```markdown
- OS: 26 beta 1
- swift: 6.2-dev
- xcode: 26
```

### Relevant log output

```shell
[...]

PR #167, opened 2025-06-24
Provides explicit makefile targets for files

### LINKED ISSUE #165: [Request]: Make target dependencies should be made explicit

madrob:
fixes #165

`init` depends on `vminitd/bin/vminitd`, `vminitd/bin/vmexec`, and `bin/cctl`
`integration` depends on `bin/containerization-integration`
`bin/*` get generated via `make containerization`


madrob:
Are there specific sections that you think are more difficult to understand? We can add some comments there.

I'm not sure what level of familiarity I can expect from folks reading the Makefile, but if the people who expect to maintain it don't understand what's going on then that's probably a good signal that we need more explanation.

dcantah (approved):

katiewasnothere (approved):
@madrob I think the changes to make it more clear what binaries are being built helps a lot! Thanks!

madrob:
I think this might have a problem where somebody running `make containerization` won't get an updated `bin/cctl` by default now. Will need to think about how to deal with this.

katiewasnothere:
@madrob Could they not just run `all`? That reminds me we also probably want to update the CI workflows

madrob:
> That reminds me we also probably want to update the CI workflows

What would you update? It looks fine to me, the problem was that I was trying to build stuff and not paying attention to the CI workflows as an example of the correct way to do things.

katiewasnothere:
@madrob I would change this step [make containerization and docs](https://github.com/apple/containerization/blob/b626d32a4869915d916f5ad12bb92ec3630cbddf/.github/workflows/containerization-build-template.yml#L51) to call `make all` which includes init, so then we can get rid of building init [here](https://github.com/apple/containerization/blob/b626d32a4869915d916f5ad12bb92ec3630cbddf/.github/workflows/containerization-build-template.yml#L60)

madrob:
> I would change this step [make containerization and docs] to call `make all` which includes init

Do you want `all` to also make `docs`? (It does not currently.)

It looks like `init` currently runs with the 6.2 swift snapshot installed by swiftly, which I think is related to #119 and #94 - I don't want to accidentally regress on that.

katiewasnothere:
@madrob Fair enough, let's hold off on those changes then.

katiewasnothere (approved):

madrob:
Pushed a squashed commit with signature.

PR #171, opened 2025-06-26
Makefile: Fix `all` not building with local changes

dcantah:
We need to run the source build phase, otherwise we can have scenarios where a user has local changes and `make` reports:

```
make: Nothing to be done for `all'.
```


dcantah:
Ok, the same problem exists for local changes to vminitd as well. I'm inclined to revert the recent change to the makefile and re-explore how we can improve it if we want to.

PR #173, opened 2025-06-26
mount /dev/console

### LINKED ISSUE #145: [Bug]: Need to set up /dev/console
### I have done the following

### Steps to reproduce

We don't set up /dev/console today in containers as a bind mount to the currently opened pty.

### Current behavior

Asking for terminal=true should cause our pal /dev/console to appear.

### Expected Behavior

/dev/console to exist after asking for a terminal.

### Environment

```markdown
- OS: 26 beta
- Swift: 6.2
- Xcode: 26 beta
```

### Relevant log output

```shell

```

### Code of Conduct

elijah-wright:
fixes #145

this PR changes `configureConsole()` to bind mount `/dev/console` if `process.terminal` is true. I couldn't think of a better way to get the pty path so I used `readlink`. most OCI images don't have `/dev/console` because they don't have `mknod` and `devtmpfs` won't create `/dev/console` without a physical console in the namespace, so it also creates `/dev/console`

@dcantah

### DIFF
--- a/vminitd/Sources/vmexec/ExecCommand.swift
+++ b/vminitd/Sources/vmexec/ExecCommand.swift
@@ -21,6 +21,10 @@ import LCShim
 import Logging
 import Musl
 
+#if canImport(Glibc)
+import Glibc
+#endif
+
 struct ExecCommand: ParsableCommand {
     static let configuration = CommandConfiguration(
         commandName: "exec",
@@ -75,6 +79,7 @@ struct ExecCommand: ParsableCommand {
 
         let childPipe = Pipe()
         try childPipe.setCloexec()
+        var hostFd: Int32 = -1
         let processID = fork()
 
         guard processID != -1 else {
@@ -93,11 +98,45 @@ struct ExecCommand: ParsableCommand {
                 throw App.Errno(stage: "setsid()")
             }
 
+            if process.terminal {
+                var containerFd: Int32 = 0
+                var ws = winsize(ws_row: 40, ws_col: 120, ws_xpixel: 0, ws_ypixel: 0)
+                guard openpty(&hostFd, &containerFd, nil, nil, &ws) == 0 else {
+                    throw App.Errno(stage: "openpty()")
+                }
+
+                guard dup3(containerFd, 0, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stdin)")
+                }
+                guard dup3(containerFd, 1, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stdout)")
+                }
+                guard dup3(containerFd, 2, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stderr)")
+                }
+                _ = close(containerFd)
+
+                guard ioctl(0, UInt(TIOCSCTTY), 0) != -1 else {
+                    throw App.Errno(stage: "setctty()")
+                }
+            }
+
+            var fdCopy = hostFd
+            let fdData = Data(bytes: &fdCopy, count: MemoryLayout.size(ofValue: fdCopy))
+            try childPipe.fileHandleForWriting.write(contentsOf: fdData)
+            try childPipe.fileHandleForWriting.close()
+
+            var ackBuf: UInt8 = 0
+            _ = read(4, &ackBuf, 1)
+            close(4)
+            close(hostFd)
+
             // Apply O_CLOEXEC to all file descriptors except stdio.
             // This ensures that all unwanted fds we may have accidentally
             // inherited are marked close-on-exec so they stay out of the
             // container.
             try App.applyCloseExecOnFDs()
+
             try App.setRLimits(rlimits: process.rlimits)
 
             // Change stdio to be owned by the requested user.
@@ -105,12 +144,6 @@ struct ExecCommand: ParsableCommand {
 
             // Set uid, gid, and supplementary groups
             try App.setPermissions(user: process.user)
-
-            if process.terminal {
-                guard ioctl(0, UInt(TIOCSCTTY), 0) != -1 else {
-                    throw App.Errno(stage: "setctty()")
-                }
-            }
 
             try App.exec(process: process)
         } else {  // parent process
@@ -120,9 +153,9 @@ struct ExecCommand: ParsableCommand {
             _ = try childPipe.fileHandleForReading.readToEnd()
             try childPipe.fileHandleForReading.close()
 
-            // send our child's pid to our parent before we exit.
-            var childPid = processID
-            let data = Data(bytes: &childPid, count: MemoryLayout.size(ofValue: childPid))
+            // send our child's pid and the host fd to our parent before we exit.
+            var payload = [Int32(processID), hostFd]
+            let data = Data(bytes: &payload, count: MemoryLayout<Int32>.size * 2)
 
             try syncfd.write(contentsOf: data)
             try syncfd.close()
--- a/vminitd/Sources/vmexec/RunCommand.swift
+++ b/vminitd/Sources/vmexec/RunCommand.swift
@@ -21,6 +21,14 @@ import LCShim
 import Logging
 import Musl
 
+#if canImport(Musl)
+import Musl
+private let _ptsname = Musl.ptsname
+#elseif canImport(Glibc)
+import Glibc
+private let _ptsname = Glibc.ptsname
+#endif
+
 struct RunCommand: ParsableCommand {
     static let configuration = CommandConfiguration(
         commandName: "run",
@@ -68,6 +76,7 @@ struct RunCommand: ParsableCommand {
 
         let childPipe = Pipe()
         try childPipe.setCloexec()
+        var hostFd: Int32 = -1
         let processID = fork()
 
         guard processID != -1 else {
@@ -92,6 +101,35 @@ struct RunCommand: ParsableCommand {
 
             try childRootSetup(rootfs: root, mounts: spec.mounts, log: log)
 
+            if process.terminal {
+                let containerMount = ContainerMount(rootfs: root.path, mounts: spec.mounts)
+                try containerMount.configureConsole()
+                var containerFd: Int32 = 0
+                var ws = winsize(ws_row: 40, ws_col: 120, ws_xpixel: 0, ws_ypixel: 0)
+                guard openpty(&hostFd, &containerFd, nil, nil, &ws) == 0 else {
+                    throw App.Errno(stage: "openpty()")
+                }
+
+                guard dup3(containerFd, 0, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stdin)")
+                }
+                guard dup3(containerFd, 1, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stdout)")
+                }
+                guard dup3(containerFd, 2, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stderr)")
+                }
+                _ = close(containerFd)
+
+                guard ioctl(0, UInt(TIOCSCTTY), 0) != -1 else {
+                    throw App.Errno(stage: "setctty()")
+                }
+
+                if let cPtr = _ptsname(hostFd) {
+                    mountConsole(path: String(cString: cPtr))
+                }
+            }
+
             if !spec.hostname.isEmpty {
                 let errCode = spec.hostname.withCString { ptr in
                     Musl.sethostname(ptr, spec.hostname.count)
@@ -100,6 +138,16 @@ struct RunCommand: ParsableCommand {
                     throw App.Errno(stage: "sethostname()")
                 }
             }
+
+            var fdCopy = hostFd
+            var fdData = Data(bytes: &fdCopy, count: MemoryLayout.size(ofValue: fdCopy))
+            try childPipe.fileHandleForWriting.write(contentsOf: fdData)
+            try childPipe.fileHandleForWriting.close()
+
+            var ackBuf: UInt8 = 0
+            _ = read(4, &ackBuf, 1)
+            close(4)
+            close(hostFd)
 
             // Apply O_CLOEXEC to all file descriptors except stdio.
             // This ensures that all unwanted fds we may have accidentally
@@ -115,12 +163,6 @@ struct RunCommand: ParsableCommand {
             // Set uid, gid, and supplementary groups.
             try App.setPermissions(user: process.user)
 
-            if process.terminal {
-                guard ioctl(0, UInt(TIOCSCTTY), 0) != -1 else {
-                    throw App.Errno(stage: "setctty()")
-                }
-            }
-
             try App.exec(process: process)
         } else {  // parent process
             try childPipe.fileHandleForWriting.close()
@@ -129,9 +171,9 @@ struct RunCommand: ParsableCommand {
             _ = try childPipe.fileHandleForReading.readToEnd()
             try childPipe.fileHandleForReading.close()
 
-            // send our child's pid to our parent before we exit.
-            var childPid = processID
-            let data = Data(bytes: &childPid, count: MemoryLayout.size(ofValue: childPid))
+            // send our child's pid and the host fd to our parent before we exit.
+            var payload = [Int32(processID), hostFd]
+            let data = Data(bytes: &payload, count: MemoryLayout<Int32>.size * 2)
 
             try syncfd.write(contentsOf: data)
             try syncfd.close()
@@ -141,7 +183,6 @@ struct RunCommand: ParsableCommand {
     private func mountRootfs(rootfs: String, mounts: [ContainerizationOCI.Mount]) throws {
         let containerMount = ContainerMount(rootfs: rootfs, mounts: mounts)
         try containerMount.mountToRootfs()
-        try containerMount.configureConsole()
     }
 
     private func prepareRoot(rootfs: String) throws {
@@ -255,5 +296,16 @@ struct RunCommand: ParsableCommand {
         let cStringCopy = UnsafeMutableBufferPointer<CChar>.allocate(capacity: cString.count)
         _ = cStringCopy.initialize(from: cString)
         return UnsafeMutablePointer(cStringCopy.baseAddress)
+    }
+
+    private func mountConsole(path: String) {
+        let console = "/dev/console"
+        if access(console, F_OK) != 0 {
+            let fd = open(console, O_RDWR | O_CREAT, mode_t(UInt16(0o600)))
+            if fd != -1 {
+                close(fd)
+            }
+        }
+        _ = mount(path, console, "", UInt(MS_BIND), nil)
     }
 }
--- a/vminitd/Sources/vmexec/RunCommand.swift
+++ b/vminitd/Sources/vmexec/RunCommand.swift
@@ -140,7 +140,7 @@ struct RunCommand: ParsableCommand {
             }
 
             var fdCopy = hostFd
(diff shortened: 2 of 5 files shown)


dcantah:
Nice, will look at this today. The alternative approach that runc takes (although I'd need to remember why, the likely answer is security or correctness) is to make the pty in the "container" (during the setup, but definitely after all the namespaces are up), and then send the fd via a unix domain socket, although we could use pidfd as well, to the parent. That'd be a much more involved change however.

brauner:
> Nice, will look at this today. The alternative approach that runc takes (although I'd need to remember why, the likely answer is security) is to make the pty in the "container" (during the setup, but definitely after all the namespaces are up), and then send the fd via a unix domain socket, although we could use pidfd as well, to the parent. That'd be a much more involved change however.



> Nice, will look at this today. The alternative approach that runc takes (although I'd need to remember why, the likely answer is security) is to make the pty in the "container" (during the setup, but definitely after all the namespaces are up), and then send the fd via a unix domain socket, although we could use pidfd as well, to the parent. That'd be a much more involved change however.

If the pty device originated from a devpts instance outside the containers mount namespace resolving that pty will fail with ENODEV in various libc functions confusing some tools.

There's other very security relevant aspects but those only come into play once you start allocating pty FDs when spawning a shell or process in the container. Doing that correctly is a bit more involved as allocating ptys from inside the container must be done via rather involved special purpose APIs.

elijah-wright:
yeah like @brauner said I wasn't looking at that too much but I think there are some `libc` helpers that will try to resolve the path of the pty, maybe `ptsname_r` or `ttyname_r` but I'd have to look at it. the pty device is created by `TerminalIO.create()` which happens before the call to `/sbin/vmexec` or really `unshare(CLONE_NEWPID | CLONE_NEWNS | CLONE_NEWUTS)`

allocating the pty in the container would avoid working with the container's `/dev` from the runtime process and there's probably some other stuff, maybe accurate attributes. but I think it's also tricky because if you mount `devpts` with [`newinstance`](https://linuxcommand.org/lc3_man_pages/mount8.html) then you can't open `/dev/pts/*` from the host. if you have a `pidfd` I think you could use `pidfd_getfd`? if not a Unix socket with `SCM_RIGHTS` would probably work

dcantah:
>If the pty device originated from a devpts instance outside the containers mount namespace resolving that pty will fail with ENODEV in various libc functions confusing some tools.

Makes sense okay

@elijah-wright Here is what I'm mentioning https://github.com/opencontainers/runc/blob/b1722d790214952e8a20d4ddd6a83451b9b665a1/libcontainer/init_linux.go#L370. The tricky part is like you'd mentioned, the setup for IO for our process type assumes we `.start()` it before launching the process as we currently create it outside, but now specifically just for the `TerminalIO` type we need to grab the pty master fd from vmexec and *then* start the relay.

dcantah:
>if you have a pidfd I think you could use pidfd_getfd? if not a Unix socket with SCM_RIGHTS would probably work

Yep!

elijah-wright:
do you want me to change the PR to match `runc` then?

dcantah:
@elijah-wright Yes, I think it's the right approach.

elijah-wright:
@dcantah can you look at my new changes

dcantah:
@elijah-wright Will look at the new push tomorrow morning (or tonight possibly), but can you run `make fmt` and re-push to make the CI nice and happy 😄

elijah-wright:
oh ok ty. I'm at work but I can do it in a few hours

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -93,7 +98,30 @@ struct RunCommand: ParsableCommand {
...
+                }
+                guard dup3(containerFd, 1, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stdout)")
+                }
+                guard dup3(containerFd, 2, 0) != -1 else {
+                    throw App.Errno(stage: "dup3(slave->stderr)")
+                }
+                _ = close(containerFd)
+
+                guard ioctl(0, UInt(TIOCSCTTY), 0) != -1 else {
```
dcantah, line 123: We also set the ctty on l152-156 so we'll need to remove

elijah-wright: oh I think I copy and pasted the if statement on accident

vminitd/Sources/vmexec/ExecCommand.swift
```diff
@@ -96,11 +101,41 @@ struct ExecCommand: ParsableCommand {
...
+            try childPipe.fileHandleForWriting.write(contentsOf: fdData)
+            try childPipe.fileHandleForWriting.close()
+
             // Apply O_CLOEXEC to all file descriptors except stdio.
             // This ensures that all unwanted fds we may have accidentally
             // inherited are marked close-on-exec so they stay out of the
             // container.
-            try App.applyCloseExecOnFDs()
+            let preserve: Set<Int32> = hostFd >= 0 ? [hostFd] : []
+            try App.applyCloseExecOnFDs(preserve: preserve)
```
dcantah: I don't believe we want to leave this open across the exec. Ideally we have the parent "ack" the fd receive and then we can close the master here and continue on to exec.

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -104,11 +132,17 @@ struct RunCommand: ParsableCommand {
...
+            try childPipe.fileHandleForWriting.write(contentsOf: fdData)
+            try childPipe.fileHandleForWriting.close()
+
             // Apply O_CLOEXEC to all file descriptors except stdio.
             // This ensures that all unwanted fds we may have accidentally
             // inherited are marked close-on-exec so they stay out of the
             // container.
-            try App.applyCloseExecOnFDs()
+            let preserve: Set<Int32> = hostFd >= 0 ? [hostFd] : []
+            try App.applyCloseExecOnFDs(preserve: preserve)
```
dcantah: Same comment as in ExecCommand, we'll want to close this, but have some way for the parent to ack that it received the fd first and it's all good to start receiving and sending io.

vminitd/Sources/vminitd/TerminalIO.swift
```diff
@@ -36,30 +40,47 @@ final class TerminalIO: ManagedProcess.IO & Sendable {
...
+            try parent?.resize(size: size)
         }
     }
 
-    func start() throws {
+    func start() throws {}
+
+    func attach(pid: Int32, fd: Int32) throws {
+        #if os(linux)
+        let containerFd = Glibc.syscall(Int(SYS_pidfd_open), pid, 0)
```
dcantah: We only build against musl today, the entire guest userland is just two statically linked executables (vminitd and the vmexec one you've been making most of the changes in). I'm fine supporting glibc if anyone wants to craft their own userland that's more full featured, but you'll need to do something like this here https://github.com/apple/containerization/blob/1a59de82b052a86edd9e2bd6f023d212a061b0b4/Sources/ContainerizationOCI/Bundle.swift#L20-L28.

vminitd/Sources/vmexec/ExecCommand.swift
elijah-wright: if we're using an ack pipe and sending the descriptor from the `vmexec` parent then wouldn't that not guarantee that `vminitd` has duplicated it? or maybe you were thinking of sending an ack from `vminitd`

dcantah: Correct yea, we'd wait here on some pipe until we got a response

dcantah: (where "response" can just be vminitd closing the pipe)

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -93,7 +98,30 @@ struct RunCommand: ParsableCommand {
                 throw App.Errno(stage: "setsid()")
             }
 
-            try childRootSetup(rootfs: root, mounts: spec.mounts, log: log)
+            try childRootSetup(rootfs: root, mounts: spec.mounts, process: process, log: log)
```
dcantah: Wait, how does this work? Inside `childRootSetup` we do the rootfs setup dance, and one of those steps includes `configureConsole`. However, we haven't set stdio to the pty we create here yet so fd 0 should still be /dev/null I'd imagine. I think we need to move `configureConsole` to inside of the below `if process.terminal` branch as the last step. Also we don't need to do the readlink dance anymore as we can just `ptsname()` now.

vminitd/Sources/vmexec/Mount.swift
```diff
@@ -44,6 +48,29 @@ struct ContainerMount {
         guard symlink("pts/ptmx", ptmx) == 0 else {
             throw App.Errno(stage: "symlink(pts/ptmx)")
         }
+
+        if process.terminal {
+            var buf = [CChar](repeating: 0, count: 4096)
+            let len = readlink("/proc/self/fd/0", &buf, buf.count - 1)
```
dcantah: Because we create the pty inside of the namespaces now, we should just be able to ptsname the fd

elijah-wright: yeah I forgot about the original code, can we call `ptsname` in `execInNamespace` and get rid of the changes to `configureConsole`?

dcantah: We should keep it. Here's what I'd propose: leave `configureConsole` as is, and then we can have a `mountConsole` (or whatever you think is a good name) method that takes in the path to the slave side pty and then just bind mounts it over a /dev/console file we create. 

`mountConsole` would mostly be:
```
open("/dev/console", O_CREAT, ...)
mount(slavePath, "/dev/console", "", "bind", MS_BIND, "")
```

The mount should be done right after we're done setting things up (we set ctty), but prior to sending the fd over the pipe to vminitd.

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -104,23 +142,27 @@ struct RunCommand: ParsableCommand {
...
             }
 
+            var fdCopy = hostFd
+            var fdData = Data(bytes: &fdCopy, count: MemoryLayout.size(ofValue: fdCopy))
+            try childPipe.fileHandleForWriting.write(contentsOf: fdData)
+            try childPipe.fileHandleForWriting.close()
+
+            var ackBuf: UInt8 = 0
+            _ = read(4, &ackBuf, 1)
+            _ = close(4)
```
dcantah: Once this is acked we should be able to close hostFd and we can remove the preserve logic

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -133,6 +133,10 @@ extension ManagedProcess {
         try self.lock.withLock {
             log.debug("starting managed process")
 
+            if !($0.io is TerminalIO) {
+                try $0.io.start()
```
dcantah: Could we just have TerminalIO's `start()` just be a nop so you don't need any special logic here? We can just call `try $0.io.start()`

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -144,19 +148,37 @@ extension ManagedProcess {
...
+            if piddata.count >= MemoryLayout<Int32>.size * 2 {
+                fd = piddata.withUnsafeBytes { ptr in
+                    ptr.load(fromByteOffset: MemoryLayout<Int32>.size, as: Int32.self)
+                }
+            }
+
+            log.info("got back pid data \(i), fd \(fd)")
             $0.pid = i
 
+            if let terminalIO = $0.io as? TerminalIO, fd != -1 {
```
dcantah: Somewhat same thing here on trying to avoid the cast. I think a decent check would just be if the user asked for a terminal (so store the terminal bool that's on HostStdio in the constructor somewhere on this object that we can access here) and then:

```swift
if self.terminal {
    guard fd != -1 else {
         throw SomeError()
    }
    try $0.io.attach(pid: i, fd: fd)
}

```

vminitd/Sources/vminitd/ManagedProcess.swift
elijah-wright: I think it is a nop right?

dcantah: It is yep, hadn't gotten that far in the file 😆. Then we don't need any special cased logic here. It'll do nothing for TerminalIO and StandardIO will start the relays

vminitd/Sources/vminitd/ManagedProcess.swift
elijah-wright: what if we extend `attach()` to `IO` and nop it in `StandardIO`?

dcantah: I think it should be on the protocol (and if it isn't that's my bad, I vaguely remembered, or hallucinated, that you had it on the protocol at one point so didn't recheck), but I'd still prefer the branch. Honestly, now that I'm saying that I think the $0.io.start() should be guarded by a !self.terminal {} as well just for clarity/consistency.

dcantah: By the way thanks so much for the quick turnarounds!

elijah-wright:
@dcantah I think I fixed the build problems

elijah-wright:
@dcantah I'm stuck on the `syscall` import

dcantah:
@elijah-wright I'll take a look, let me pull this locally

dcantah:
@elijah-wright Seems Swift can't generate bindings for variadic c functions is the problem 😔, so I've written us some wrappers here we can use for this. This should be checked in tomorrow morning I imagine https://github.com/apple/containerization/pull/189. You're going to want to rebase here also as it seems a little behind

dcantah:
Ok you should be set to rebase on main and grab the new wrappers

elijah-wright:
@dcantah ok after some trial and error I think it's good, I did some work getting it ready for the new wrappers yesterday but rebase kept unsigning the commits

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -21,6 +21,14 @@ import LCShim
 import Logging
 import Musl
 
+#if canImport(Musl)
+import Musl
```
dcantah, line 24: Remove the musl import on l22 if we wanna do this

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -256,4 +297,15 @@ struct RunCommand: ParsableCommand {
...
+
+    private func mountConsole(path: String) {
+        let console = "/dev/console"
+        if access(console, F_OK) != 0 {
+            let fd = open(console, O_RDWR | O_CREAT, mode_t(UInt16(0o600)))
+            if fd != -1 {
+                close(fd)
+            }
+        }
+        _ = mount(path, console, "", UInt(MS_BIND), nil)
```
dcantah: Lets not ignore the error here

vminitd/Sources/vminitd/TerminalIO.swift
```diff
@@ -35,30 +36,49 @@ final class TerminalIO: ManagedProcess.IO & Sendable {
...
+
+        let hostFd = CZ_pidfd_getfd(containerFd, fd, 0)
+        guard Foundation.close(Int32(containerFd)) == 0 else {
+            self.log?.error("failed to close pidfd: \(POSIXError.fromErrno())")
+            return
+        }
+
+        guard hostFd != -1 else {
+            throw POSIXError.fromErrno()
+            return
```
dcantah, line 89: This return will never get hit

vminitd/Sources/vminitd/TerminalIO.swift
```diff
@@ -35,30 +36,49 @@ final class TerminalIO: ManagedProcess.IO & Sendable {
...
+    func attach(pid: Int32, fd: Int32) throws {
+        #if os(Linux)
+        let containerFd = CZ_pidfd_open(pid, 0)
+        guard containerFd != -1 else {
+            throw POSIXError.fromErrno()
         }
+
+        let hostFd = CZ_pidfd_getfd(containerFd, fd, 0)
+        guard Foundation.close(Int32(containerFd)) == 0 else {
+            self.log?.error("failed to close pidfd: \(POSIXError.fromErrno())")
```
dcantah: We should actually throw here, with how it is today it will silently succeed to the user

elijah-wright: I haven't looked at how `.error()` works, does it only log?

dcantah: Correct yep

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -118,11 +126,6 @@ final class ManagedProcess: Sendable {
             )
         }
 
-        log.info("starting io")
-
-        // Setup IO early. We expect the host to be listening already.
-        try io.start()
```
dcantah, line 124: I'm seeing one large blocker while trying this locally (and please do the same, `make && make integration` to run the tests before pushing any updates). The way we've gone about IO so far is the contract is the guest should dial the host vsock ports before `start()` on the process is called. The rationale being that we want valid descriptors before the process is launched so we don't possibly miss any data. This moves the start of the io to `start()` itself which I didn't really put together in my head, so the container never actually comes up as we error out on the host after 3 seconds waiting for the guest to dial. I'll have to play around with this locally to see what we can do..

elijah-wright: I think we can revert that right?

dcantah: The problem is `start` for the TerminalIO is a nop now, and we rely on the sockets being dialed at the very least or we'll timeout on the host. I'm still digging a bit, but I think what we can do is dial the sockets in `start`, but just not setup the relays until we get back the slave pty handle and then call `attach()`. That should (emphasis on should 🤣) work I'd imagine.

elijah-wright: oh yeah right, I think it would work. I think that's what containerd does only with fifos rather than sockets

elijah-wright: @dcantah do you want me to move the stuff in `setupRelays()` to `start()` and try `io.start()` in `init()`?

dcantah: I'm playing around in this area right now, if I don't have a decent approach tomorrow morning please do. I'm trying not to cause too much churn in the current IO setup, even though I'm not a huge fan of what we have in main today

elijah-wright:
@dcantah I haven't looked at your changes yet, do you want to implement them here? I can do it in a few hours if not

dcantah:
Yea I can fool around with it and post a patch we can apply here

dcantah:
Sorry for delay, the long weekend consumed me 😂. I have this almost functioning and will push a patch here you can apply and we can work together on

elijah-wright:
no that's ok! do you want me to squash these commits together and then you can apply the patch yourself? if not that's ok. I want you to get credit for what you did

dcantah:
Squashing now would help, but we can just put a "co-authored by" line on the final commit so it doesn't matter much

dcantah:
@elijah-wright I ended up posting a separate PR with you as a co-author here https://github.com/apple/containerization/pull/248. I realized we needed quite a bit of reworking on the IO types to get systemd happy (which this change was aimed at originally) which made the patches diverge so much that review would be confusing here. Thanks so much for working on this so far!

elijah-wright:
yea np! I think I learned a lot from this lol. do you want me to close this?

dcantah:
Yea I can close it out. I couldn't add you as a reviewer over on #248 but please do review if you want to, curious on your insight!

PR #174, opened 2025-06-26
vmexec: Set perms on stdio

dcantah:
We don't chown the container's stdio today.

### DIFF
--- a/vminitd/Sources/vmexec/vmexec.swift
+++ b/vminitd/Sources/vmexec/vmexec.swift
@@ -106,6 +106,24 @@ extension App {
         }
     }
 
+    static func fixStdioPerms(user: ContainerizationOCI.User) throws {
+        for i in 0...2 {
+            var fdStat = stat()
+            try withUnsafeMutablePointer(to: &fdStat) { pointer in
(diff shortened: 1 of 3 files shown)


vminitd/Sources/vmexec/vmexec.swift
```diff
@@ -106,6 +106,24 @@ extension App {
...
+            var fdStat = stat()
+            try withUnsafeMutablePointer(to: &fdStat) { pointer in
+                guard fstat(Int32(i), pointer) == 0 else {
+                    throw App.Errno(stage: "fstat(fd)")
+                }
+            }
+
+            let desired = uid_t(user.uid)
+            if fdStat.st_uid != desired {
+                guard fchown(Int32(i), desired, fdStat.st_gid) != -1 else {
```
crosbymichael, line 120: Should gid be set to user.gid?

dcantah: I vaguely remember it's not needed.. but I cannot remember why

dcantah: I believe I'm thinking of runc's impl, but crun seems to use the passed in gid.

PR #189, opened 2025-07-01
LCShim: Add CZ prefixes/pidfd wrappers

dcantah:
We should namespace these a bit, and we need pidfd for some upcoming terminal work. This also just gets rid of syscall2 and replaces with a pivot_root wrapper.

### DIFF
--- a/vminitd/Sources/LCShim/syscall.c
+++ b/vminitd/Sources/LCShim/syscall.c
@@ -14,13 +14,24 @@
  * limitations under the License.
  */
 
-//
+#include <sys/prctl.h>
+#include <sys/syscall.h>
+#include <unistd.h>
 
-#ifndef __SYSCALL2_H
-#define __SYSCALL2_H
+#include "syscall.h"
 
-int syscall2(long number, void *arg1, void *arg2);
+int CZ_pivot_root(const char *new_root, const char *put_old) {
+  return syscall(SYS_pivot_root, new_root, put_old);
+}
 
-int set_sub_reaper();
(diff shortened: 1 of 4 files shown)


vminitd/Sources/LCShim/syscall.c
```diff
@@ -18,10 +18,20 @@
...
-int syscall2(long number, void *arg1, void *arg2) {
-  return syscall(number, arg1, arg2);
+int CZ_pivot_root(const char *new_root, const char *put_old) {
+  return syscall(SYS_pivot_root, new_root, put_old);
 }
 
-int set_sub_reaper() { return prctl(PR_SET_CHILD_SUBREAPER, 1); }
+int CZ_set_sub_reaper() { return prctl(PR_SET_CHILD_SUBREAPER, 1); }
+
+int CZ_pidfd_open(pid_t pid, unsigned int flags) {
```
jglogan, line 29: Do we need to define these functions for musl builds if they don't exist there? Or is the idea that we want syscall to just ENOSYS these guys?

dcantah: pidfd_open has been in since 5.6 iirc which was 5 years ago now. I'm somewhat alright drawing a line there and saying that is the minimum kernel version supported. We should formalize this though in a doc here. I'm trying to think of any other feature we use that might require an even newer kernel.

dcantah: But I'm fine ENOSYS'ing yes

PR #193, opened 2025-07-03
Replace `NSLock` with `Mutex`

dkovba:
This PR replaces `NSLock` with `Mutex`. Thanks @dcantah for the idea!


dcantah:
Nice, this doesn't see the issue you'd messaged about anymore?

dcantah (approved):

dkovba:
> Nice, this doesn't see the issue you'd messaged about anymore?

Confirmed that this implementation doesn't have the issue we had with the previous implementation using a `Mutex` and a struct.

PR #194, opened 2025-07-03
Vminitd: Rework ManagedProcess.IO protocols

dcantah:
I think the way these types were structured made reasoning about them a bit difficult. I don't think this fully solves the problem, but this change aims to make things a bit simpler by hoisting a lot of the logic for the IO relays to a new IOPair type thats goal is to simply take in a reader and writer and handle their resource cleanup after a relay finishes. Bundled in with this is also the buffer we'll use to copy between them.

This change:

- Alters ManagedProcess.IO `start` to take in the process to alter instead of this just being a side effect of the constructors.
- Gets rid of `close()` in favor of `CloseStdin`. The IO will get closed when the relays finish, which will naturally happen if the process exits or just closes its side of the pipes/pty. This makes it so that the one special case (a client wants to signal no more input is coming) is still sane.
- Move all relay logic and resource cleanup to a new IOPair type that takes in protocols that are easily conformable by all of our various io types ( Socket, Terminal, FileHandle).

### DIFF
--- a/vminitd/Sources/vminitd/IOCloser.swift
+++ b/vminitd/Sources/vminitd/IOCloser.swift
@@ -1,0 +1,21 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+protocol IOCloser: Sendable {
+    var fileDescriptor: Int32 { get }
+
+    func close() throws
+}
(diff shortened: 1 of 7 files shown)


dcantah:
@elijah-wright Here is what I was mentioning. I think this should hopefully make things less confusing with these types. For TerminalIO.start you can just dial the sockets and store them, and then in your new attach method you can create these new IOPair types and just call relay on them. I'll wait to see if the others hate this 😄

vminitd/Sources/vminitd/IOCloser.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+protocol IOCloser: Sendable {
+    var fileDescriptor: Int32 { get }
```
crosbymichael, line 18: Why does a closer need the fd exposed?

crosbymichael (approved):
LGTM with one question on the protocol.

vminitd/Sources/vminitd/IOCloser.swift
dcantah: Because I can't name things for the life of me 🙂. Do you have a better name for the IOCloser? It's moreso a type that lets us do IO against that has a way to clean itself up, so just closer is a bit odd I agree

dcantah: I'm going to check this in as a find and replace is simple enough in the future and this should help our friend Elijah with his terminal work

PR #198, opened 2025-07-07
Add a `com.apple.containerization.index.indirect` annotation to distinguish synthesized index

stevapple:
In general I believe it's clever to synthesize an index for single-platform image manifest, but we still need a way to distinguish it. Add a dedicated annotation is the slightest change I've come up with, and it's also OCI compliant. With this change come in, we can work around https://github.com/apple/container/issues/212 and imitate the behavior of other runtime with `container`.

Note that since `cctl` is meant to be a dedicated tool for inspecting the Containerization framework itself, I didn't apply the indirection for it, and it will be as-is with the genuine storage.

### DIFF
--- a/Sources/Containerization/Image/ImageStore/ImageStore+Import.swift
+++ b/Sources/Containerization/Image/ImageStore/ImageStore+Import.swift
@@ -211,10 +211,12 @@ extension ImageStore {
                 let platform = supportedPlatforms.first!
                 var root = root
                 root.platform = platform
-                let index = ContainerizationOCI.Index(schemaVersion: 2, manifests: [root], annotations: [
-                    // indicate that this is a synthesized index which is not directly user facing
-                    AnnotationKeys.containerizationIndexIndirect: "true"
-                ])
+                let index = ContainerizationOCI.Index(
+                    schemaVersion: 2, manifests: [root],
+                    annotations: [
+                        // indicate that this is a synthesized index which is not directly user facing
+                        AnnotationKeys.containerizationIndexIndirect: "true"
+                    ])
                 return index
             default:
                 throw ContainerizationError(.internalError, message: "Failed to create index for descriptor \(root.digest), media type \(root.mediaType)")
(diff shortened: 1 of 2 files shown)


adityaramani:
Thanks for the PR! Have you run into a case where using the synthesized index has broken other container image tools? 

I think adding that annotation makes sense, but we probably dont need the additional methods for obtaining the original manifest / descriptor here in Containerization? The consumers of the library (in this case container) can check for the annotation and make decisions on how to handle it?

What do you think? Also, we have enable commit verification for this repository, could please set-up commit signing and re-push your commits?

stevapple:
> Thanks for the PR! Have you run into a case where using the synthesized index has broken other container image tools?

Without the functionality, we'll not be able to pass [`critest`](https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/validation.md). The result shows that returning the original manifest (that matches the digest) is expected.

> I think adding that annotation makes sense, but we probably dont need the additional methods for obtaining the original manifest / descriptor here in Containerization? The consumers of the library (in this case container) can check for the annotation and make decisions on how to handle it?
>
> What do you think? 

It makes sense. These codes exist because the indirection was once applied to `cctl`, and I didn't remove it when I cleaned up the `cctl` part.

> Also, we have enable commit verification for this repository, could please set-up commit signing and re-push your commits?

Thanks for your advice, I'll look into that later.

Sources/Containerization/Image/ImageStore/ImageStore+Import.swift
```diff
@@ -211,7 +211,10 @@ extension ImageStore {
                 let platform = supportedPlatforms.first!
                 var root = root
                 root.platform = platform
-                let index = ContainerizationOCI.Index(schemaVersion: 2, manifests: [root])
+                let index = ContainerizationOCI.Index(schemaVersion: 2, manifests: [root], annotations: [
+                    // indicate that this is a synthesized index which is not directly user facing
+                    "com.apple.containerization.index.indirect" : "true"
```
adityaramani: Can we add this to the `Annotations` enum we have in ContainerizationOCI like you had in your previous diff?

adityaramani (changes requested):

Sources/Containerization/Image/ImageStore/ImageStore+Import.swift
stevapple: Done:)

adityaramani:
Whoops. Looks like you need to run `make fmt` and re-push

PR #199, opened 2025-07-07
Handle spaces in current dir by quoting paths

Singletoned:
Added "" around some paths so that spaces in directory names will work


dcantah:
@Singletoned LGTM, and thank you! We require commit signatures however, do you mind signing (ssh/gpg etc.) and then repushing?

Singletoned:
I've added an SSH signature and reputed.  Thanks for giving me the nudge to actually set that up.

PR #200, opened 2025-07-07
Add ContainerManager

crosbymichael:
A ContainerManager is a type that handles more of the required resource needed to create and run a container.


Sources/Containerization/ContainerManager.swift
```diff
@@ -0,0 +1,165 @@
...
+            initialFilesystem: initfs,
+            bootlog: self.imageStore.path.appendingPathComponent("bootlog.log").absolutePath()
+        )
+        try Self.createRootDirectory(path: self.imageStore.path)
+    }
+
+    /// Create a new manager with the provided kernel and image reference for the initfs.
+    public init(
+        kernel: Kernel,
+        initfsReference: String
```
dcantah: We might wanna take in an optional boot log path

dcantah: Although honestly the point of this type is it's mostly opinionated, they could always set things up themselves

crosbymichael: It sets up the bootlog inside the type, into the root directory.

Sources/Containerization/Image/ImageStore/ImageStore.swift
```diff
@@ -22,29 +22,66 @@ import Foundation
...
+        }
+    }
+
+    private static func defaultRoot() -> URL {
+        let root = FileManager.default.urls(
+            for: .applicationSupportDirectory,
+            in: .userDomainMask
+        ).first
+        guard let root else {
+            fatalError("unable to get Application Support directory for current user")
```
dcantah: I'm usually not a huge fan of panic'ing in libraries, but this one seems somewhat harmless. However, this is private and the place we call it from also is marked as throws, so can we just bubble up an error instead?

Sources/Containerization/ContainerManager.swift
```diff
@@ -0,0 +1,165 @@
...
+import ContainerizationOCI
+import ContainerizationOS
+import Foundation
+
+import Crypto
+
+extension String {
+    fileprivate func hash() throws -> String {
+        guard let data = self.data(using: .utf8) else {
+            fatalError("\(self) could not be converted to Data")
```
dcantah: Same here. This is marked as throws, and isn't public where we'd prefer the convenience of maybe not wanting to do a `try whatever()` so can we just throw some err

Sources/Containerization/ContainerManager.swift
```diff
@@ -0,0 +1,165 @@
...
+    ) async throws -> LinuxContainer {
+        let image = try await imageStore.get(reference: reference, pull: true)
+        return try await create(id, image: image, rootfsSize: rootfsSize)
+    }
+
+    /// Create a new container from the provided image.
+    public func create(
+        _ id: String,
+        image: Image,
+        rootfsSize: UInt64 = 8.gib()
```
dcantah:
```suggestion
        rootfsSizeInBytes: UInt64 = 8.gib()
```

Sources/Containerization/ContainerManager.swift
```diff
@@ -0,0 +1,165 @@
...
+            at: path.appendingPathComponent("containers"),
+            withIntermediateDirectories: true
+        )
+    }
+
+    /// Create a new container from the provided image reference.
+    public func create(
+        _ id: String,
+        reference: String,
+        rootfsSize: UInt64 = 8.gib()
```
dcantah:
```suggestion
        rootfsSizeInBytes: UInt64 = 8.gib()
```

Sources/Containerization/ContainerManager.swift
dcantah: Yea that's why I'm honestly okay with this now, as the whole point of it is that everything gets shoved in the dir and can be blown away as one unit

Sources/Containerization/Image/ImageStore/ImageStore.swift
```diff
@@ -22,29 +22,66 @@ import Foundation
 /// An ImageStore handles the mappings between an image's
 /// reference and the underlying descriptor inside of a content store.
 public actor ImageStore: Sendable {
+    /// The ImageStore path it was created with.
+    public nonisolated let path: URL
```
dcantah, line 26: My Swift brain isn't functioning, why do we need this to be nonisolated?

dcantah (approved):

Sources/Containerization/Image/ImageStore/ImageStore.swift
```diff
@@ -22,29 +22,66 @@ import Foundation
...
+        } else {
+            self.contentStore = try LocalContentStore(path: path.appendingPathComponent("content"))
+        }
+
+        self.path = path
         self.referenceManager = try ReferenceManager(path: path)
     }
+
+    /// Return the default image store for the current user.
+    public static var `default`: ImageStore {
```
adityaramani: Dont know if it matters but should this be 

```
public static let `default`: ImageState : {
     ...
} ()
```

So that multiple calls to `default` return the same object?

Sources/Containerization/Image/ImageStore/ImageStore.swift
```diff
@@ -22,29 +22,66 @@ import Foundation
...
     /// - Parameters:
     ///   - reference: Name of the image.
+    ///   - pull: Pull the image if it is not found.
     ///
     /// - Returns: A `Containerization.Image`  object whose `reference` matches the given string.
     ///   This  method throws a `ContainerizationError(code: .notFound)` if the provided reference does not exist in the `ImageStore`.
-    public func get(reference: String) async throws -> Image {
-        let desc = try await self.referenceManager.get(reference: reference)
-        return Image(description: desc, contentStore: self.contentStore)
+    public func get(reference: String, pull: Bool = false) async throws -> Image {
```
adityaramani, line 77: Should we define a `fetch` method that does this get/pull operation? Just to keep the `get` API simple ?

```
    public func fetch(reference: String) async throws -> Containerization.Image {
        do {
            return try await self.image.get(reference: reference)
        } catch let error as ContainerizationError {
            if error.code == .notFound {
                return try await self.image.pull(reference: reference)
            }
            throw error
        }
    }
```

crosbymichael: We could do that or align with containerd on the verbiage of this, I'll have to see if we can do it in a backwards compatible way.

In containerd `fetch` means only pull the blobs, `pull` means fetch+unpack. There isn't an optimized `get` that I know of.

Sources/Containerization/Image/ImageStore/ImageStore.swift
crosbymichael: Good idea since this type is an actor.

Sources/Containerization/ContainerManager.swift
```diff
@@ -0,0 +1,375 @@
...
+        self.imageStore.path.appendingPathComponent("containers")
+    }
+
+    public protocol Network: Sendable {
+        func create(_ id: String) throws -> Interface?
+        func release(_ id: String) throws
+    }
+
+    /// A network backed by vmnet on macOS.
+    @available(macOS 16.0, *)
```
dcantah: Is this supposed to be 26 now?

Sources/Containerization/ContainerManager.swift
```diff
@@ -0,0 +1,375 @@
...
+import ContainerizationOCI
+import ContainerizationOS
+import Foundation
+import SendableProperty
+import ContainerizationExtras
+import Virtualization
+import vmnet
+
+/// A type that manages containers and their resources within a root directory.
+public struct ContainerManager: Sendable {
```
dcantah, line 29: Is the new guidance going to be for mostly everyone to use this type going forward? I'm debating whether it makes sense for the network protocol and types to be outside of this type.

dcantah (approved):

Sources/Containerization/ContainerManager.swift
crosbymichael: I don't think so. It's more of a helper if you want a more unified type to work with.

Sources/Containerization/Image/ImageStore/ImageStore.swift
crosbymichael: It's a `let` but it would make you `await store.path` if it was isolated. 🤷

dcantah: Fair, that seems... odd

Sources/Containerization/ContainerManager.swift
crosbymichael: Yes, I'll change this.

PR #201, opened 2025-07-08
LinuxContainer: Add ability to close stdin

dcantah:
This allows the user the ability to raise an EOF for the containers stdin. Today there's no way to close stdin so something as simple as "cat" and relying on EOF to move the process forward doesn't work

### DIFF
--- a/Sources/Containerization/SandboxContext/SandboxContext.proto
+++ b/Sources/Containerization/SandboxContext/SandboxContext.proto
@@ -34,6 +34,8 @@ service SandboxContext {
   // Resize the tty of a given process. This will error if the process does
   // not have a pty allocated.
   rpc ResizeProcess(ResizeProcessRequest) returns (ResizeProcessResponse);
+  // Close IO for a given process.
+  rpc CloseProcessStdin(CloseProcessStdinRequest) returns (CloseProcessStdinResponse);
 
   // Proxy a vsock port to a unix domain socket in the guest, or vice versa.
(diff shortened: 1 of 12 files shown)


Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -34,6 +34,8 @@ service SandboxContext {
   // Resize the tty of a given process. This will error if the process does
   // not have a pty allocated.
   rpc ResizeProcess(ResizeProcessRequest) returns (ResizeProcessResponse);
+  // Close IO for a given process.
+  rpc CloseProcessStdin(CloseProcessStdinRequest) returns (CloseProcessStdinResponse);
```
crosbymichael, line 38: Could this be a more generic API? Something like `CloseFD` and you provide it an fd number to close?

dcantah: The problem is the most important fd to close here is the guests socket for stdin and the host would have no clue what the number is

crosbymichael: the socket is different but should the client know fd 0,1,2?  If that would be too much, maybe a bool allowing any fd to be closed via an enum?  `stdin|out|err`

dcantah: We chatted offline, but this should be fine as is

crosbymichael (approved):

PR #204, opened 2025-07-08
Make MTU configurable for Netlink

realrajaryan:
Here we add the implementation to make MTU value configurable instead of hardcoding in `linkSet`.

### DIFF
--- a/Sources/Containerization/Agent/Vminitd.swift
+++ b/Sources/Containerization/Agent/Vminitd.swift
@@ -198,10 +198,11 @@ extension Vminitd: VirtualMachineAgent {
         _ = try await client.deleteProcess(request)
     }
 
-    public func up(name: String) async throws {
+    public func up(name: String, mtu: UInt32? = nil) async throws {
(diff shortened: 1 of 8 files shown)


Sources/Containerization/Agent/Vminitd.swift
```diff
@@ -198,7 +198,7 @@ extension Vminitd: VirtualMachineAgent {
         _ = try await client.deleteProcess(request)
     }
 
-    public func up(name: String) async throws {
+    public func up(name: String, mtu: UInt32? = nil) async throws {
```
jglogan, line 201: Shouldn't `mtu` be assigned in the request here?

jglogan (changes requested):

Sources/Containerization/Agent/Vminitd.swift
jglogan: Also looks like `make fmt` is needed for a clean build in CI.

dcantah:
@realrajaryan Can you squash the commits (it's complaining the earlier commits aren't signed)

jglogan (approved):

PR #207, opened 2025-07-10
LinuxContainer: Add /etc/hosts writing functionality

### LINKED ISSUE #206: [Request]: Support writing /etc/hosts

This is a pretty commonly setup file for containers, and it should be fairly analogous to what we do with /etc/resolv.conf today already. 

### Code of Conduct

dcantah:
Closes #206 

Much like we have support for supplying DNS configurations and writing out /etc/resolv.conf, this adds a way to write out /etc/hosts for a given container.


dcantah:
Related to https://github.com/apple/container/issues/314

dcantah:
@jglogan Curious on your thoughts here. I think it was worth to not have this just be an embedded field on DNS and be a separate setup as I can see people only wanting to setup one or the other and not both. I'm also thinking whether we should just have a `writeFile(dataBlob, location: location)` style rpc for vminitd for things like this. Both resolv.conf and hosts can be serviced by just sending a byte blob of the contents to the guest instead of copying the strings and then writing one specific file in the guest.

Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -239,6 +241,19 @@ message ConfigureDnsRequest {
 
 message ConfigureDnsResponse {}
 
+message ConfigureHostsRequest {
```
jglogan, line 244: do we want a specialized API for this, or perhaps an API to install an arbitrary file?

dcantah: I know we talked on this, but I'm fine with this approach for now after thinking more. I think once the `copy` rpc is a thing we can special purpose that if the thing being copied is a single file and skip the tarball part and just transfer the bytes for the single file directly. We can just remove this rpc if that does come to fruition.

jglogan (approved):

PR #212, opened 2025-07-12
Check for supported types in the `@SendableProperty` macro

dkovba:
- The `@SendableProperty` macro now checks for supported types to prevent a misuse with unsupported types like structs, dictionaries, and arrays
- The new `@SendablePropertyUnchecked` can be used with classes and enums

### DIFF
--- a/Sources/SendablePropertyMacros/SendablePropertyMacro.swift
+++ b/Sources/SendablePropertyMacros/SendablePropertyMacro.swift
@@ -21,19 +21,20 @@ import SwiftSyntax
 import SwiftSyntaxBuilder
 import SwiftSyntaxMacros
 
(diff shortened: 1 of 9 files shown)


Sources/SendablePropertyMacros/SendablePropertyMacro.swift
```diff
@@ -21,10 +21,36 @@ import SwiftSyntax
...
+            let binding = varDecl.bindings.first,
+            let typeAnnotation = binding.typeAnnotation,
+            let id = typeAnnotation.type.as(IdentifierTypeSyntax.self)
+        else {
+            // Nothing to check.
+            return
+        }
+
+        var typeName = id.name.text
+        // Allow optionals of the allowed types.
```
dkovba, line 42: This was improved in https://github.com/apple/containerization/pull/216.

PR #214, opened 2025-07-12
KeychainHelper: Simplify lookup

dcantah:
I believe .get() does the equivalent of .exists() today internally, it just also asks for returned data if it exists, so lookup() might be a little redundant.


dcantah:
@adityaramani Tell me if this is faulty in anyway, just from eyeing the API it seems like it should function

adityaramani (approved):
I think this is fine - I dont remember why we had two different methods :/

PR #216, opened 2025-07-18
Improve a check for supported types in the macro

dkovba:
The changes in this PR improve the check for supported types in the macro.


crosbymichael:
Let's close this as we are looking to move away from the property wrapper.

PR #217, opened 2025-07-18
LinuxContainer: Support pause/resume

dcantah:
This allows pausing the (really VM) container and any processes running inside.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -154,6 +154,12 @@ public final class LinuxContainer: Container, Sendable {
         case stopped
         /// An error occurred during the lifetime of this class.
         case errored(Swift.Error)
+        /// The container is being paused.
+        case pausing(PausingState)
+        /// The container is paused.
+        case paused(PausedState)
+        /// The container is being resumed.
+        case resuming(ResumingState)
 
         struct CreatingState: Sendable {}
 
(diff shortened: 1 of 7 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -327,8 +425,8 @@ extension LinuxContainer {
 
     /// Hostname mapping configurations for the container.
     public var hosts: Hosts? {
-        get { config.hosts }
-        set { config.hosts = newValue }
+        get { config.withLock { $0.hosts } }
+        set { config.withLock { $0.hosts = newValue } }
```
dkovba: It's unsafe to use a lock inside a computed property to access a struct (`hosts`). For example, `hosts.entries = newEntries` will call both a getter and a setter with no locking between them, causing a race with any other thread that writes to `config`.

dcantah: I know. This is how we do all config updates however. We solely did this just to make LinuxContainer sendable (yes, it was a copout). I'd love to explore a different approach, but today the second .create() is called on the container the properties are meaningless and have no effect so I'm not too worried

Sources/Containerization/TimeSyncer.swift
```diff
@@ -54,6 +60,14 @@ actor TimeSyncer: Sendable {
         }
     }
 
+    func pause() async {
+        self.paused = true
+    }
+
+    func resume() async {
```
crosbymichael, line 67: Since the synced is created `paused` are we missing this `resume()` call in the normal `start()` functions of the instance?

crosbymichael:
Left one comment and then this needs a rebase.

Sources/Containerization/TimeSyncer.swift
dcantah: We are yea. I originally had things a tad different, but didn't fix up the setup of the timesyncer. I changed to starting it in `paused = false`

PR #219, opened 2025-07-21
LinuxContainer/LinuxProcess: Rework supplying configuration

dcantah:
This is a fairly large reworking, but it gets rid of something that has plagued this since release which is the properties needing to be locked to be Sendable compliant. This was somewhat of a copout because we mostly know there's not a great deal of ways to have misused the setup today, but alas we'd need to either mark the type as `@unchecked` or just find a different route for setting the configuration. This change:

Exposes the underlying Configuration type that today only housed things that aren't on the OCI spec. I'd love to just expose the OCI spec, but we don't (and possibly never will) support everything on the spec, so exposing it to be freely modified would be a bit odd. Now everything related to the container is configured on this type, and the same goes for execs.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -27,9 +27,6 @@ import struct ContainerizationOS.Terminal
 
 /// `LinuxContainer` is an easy to use type for launching and managing the
 /// full lifecycle of a Linux container ran inside of a virtual machine.
-///
-/// NOTE: Editing the properties of `LinuxContainer` after calling `start()`
-/// have no effect.
 public final class LinuxContainer: Container, Sendable {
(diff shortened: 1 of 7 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -40,25 +37,114 @@ public final class LinuxContainer: Container, Sendable {
...
+        public var dns: DNS?
+        /// The hosts to add to /etc/hosts for the container.
+        public var hosts: Hosts?
+
+        public init() {}
+    }
+
+    /// `IOHandler` informs the container process about what should be done
+    /// for the stdio streams.
+    struct IOHandler: Sendable {
```
crosbymichael, line 133: Can this be deleted now?

PR #223, opened 2025-07-22
LinuxContainer: Remove SendableProperty usage

dcantah:
The state updates because of the way the macro works weren't actually being performed under the lock.


dkovba (approved):
👍

PR #225, opened 2025-07-22
Transition to actors

dkovba:
Due to the reduced use of the macro, we can now fully transition to `Mutex`.


vminitd/Sources/vminitd/VsockProxy.swift
```diff
@@ -54,14 +54,13 @@ final class VsockProxy: Sendable {
     private let udsPerms: UInt32?
     private let log: Logger?
 
-    @SendableProperty
-    private var listener: Socket?
+    private let listener = Mutex<Socket?>(nil)
     private let task = Mutex<Task<(), Never>?>(nil)
```
dkovba: @crosbymichael Would you prefer moving `listener` and `task` inside a single `Mutex` called `state`?

crosbymichael: Yes, I think that would help on readability with the mutex changes.

dcantah: Agree, make it one lock for the two fields (plop them in a struct)

vminitd/Sources/vminitd/VsockProxy.swift
```diff
@@ -54,13 +56,18 @@ final class VsockProxy: Sendable {
...
+        var listener: Socket?
+        var task: Task<(), Never>?
+    }
+
+    private let state: Mutex<State>
 }
 
 extension VsockProxy {
     func close() throws {
+        let (listener, task) = state.withLock { ($0.listener, $0.task) }
```
crosbymichael: shouldn't all of this function happen inside of the lock? You are not protecting around a double close by doing this.

vminitd/Sources/vminitd/VsockProxy.swift
```diff
@@ -99,7 +105,7 @@ extension VsockProxy {
         )
         let uds = try Socket(type: type)
         try uds.listen()
-        listener = uds
+        state.withLock { $0.listener = uds }
```
crosbymichael: Must of the functions in this type do not look thread safe.  There is not much protecting double entry to these functions, only protection around setting the state values.

vminitd/Sources/vminitd/VsockProxy.swift
dcantah: Yea, this is one of those times where the mutex was honestly a formality to appease the compiler. I don't really think VsockProxy *needs* to be sendable, as this is the only way one can be stopped today:

```swift
func stopVsockProxy(
        request: Com_Apple_Containerization_Sandbox_V3_StopVsockProxyRequest,
        context: GRPC.GRPCAsyncServerCallContext
    ) async throws -> Com_Apple_Containerization_Sandbox_V3_StopVsockProxyResponse {
        log.debug(
            "stopVsockProxy",
            metadata: [
                "id": "\(request.id)"
            ])

        do {
            let proxy = try await state.remove(proxy: request.id)
            try proxy.close()
```

state in the above is an actor, so the second it's returned no one else calling the stop rpc can actually double close() it.

dcantah: This is start()'s logic as well:

```swift
         let proxy = VsockProxy(
                id: request.id,
                action: request.action == .into ? .dial : .listen,
                port: request.vsockPort,
                path: URL(fileURLWithPath: request.guestPath),
                udsPerms: request.guestSocketPermissions,
                log: log
            )

            try proxy.start()
            try await state.add(proxy: proxy)
```

This can be reordered to store the proxy first (which will fail if the id is already in the store. And THEN start the proxy. We can just remove from the store if the start fails.

dcantah: Alternatively, because there's only two methods that are callable (e.g. that aren't private), and they aren't async so the annoyance of reentrancy are avoided, we could just make these actors if you really wanted to check "hey has this thing been closed or started already" by checking if something is nil or otherwise.

dkovba: Made `VsockProxy` an actor to resolve this.

vminitd/Sources/vminitd/VsockProxy.swift
dkovba: `VsockProxy` is now an actor.

vminitd/Sources/vminitd/VsockProxy.swift
```diff
@@ -70,7 +68,6 @@ extension VsockProxy {
         if fm.fileExists(atPath: self.path.path) {
             try FileManager.default.removeItem(at: self.path)
         }
-        let task = task.withLock { $0 }
         task?.cancel()
```
dcantah, line 71: Can we set listener to nil at the end here, as that'd be what would present double close today even though we established it can't occur today (the guard let listener check at the top would just have us return immediately)

dcantah: Likewise, in start() we can check if listener is nil, and if not it's already been started.

dkovba: Done

dcantah (approved):

PR #228, opened 2025-07-24
ContainerManager: Make root/imagestore suppliable

dcantah:
We should be able to supply an ImageStore if the user wants a different ContentStore, or a different path.

### DIFF
--- a/Sources/Containerization/ContainerManager.swift
+++ b/Sources/Containerization/ContainerManager.swift
@@ -192,15 +192,76 @@ public struct ContainerManager: Sendable {
         }
     }
 
-    /// Create a new manager with the provided kernel and initfs mount.
+    /// Create a new manager with the provided kernel, initfs mount, image store
+    /// and optional network implementation.
     public init(
         kernel: Kernel,
         initfs: Mount,
+        imageStore: ImageStore,
         network: Network? = nil
     ) throws {
-        self.imageStore = ImageStore.default
+        self.imageStore = imageStore
         self.network = network
         try Self.createRootDirectory(path: self.imageStore.path)
+        self.vmm = VZVirtualMachineManager(
+            kernel: kernel,
+            initialFilesystem: initfs,
+            bootlog: self.imageStore.path.appendingPathComponent("bootlog.log").absolutePath()
+        )
+    }
+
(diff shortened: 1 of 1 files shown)


Sources/Containerization/ContainerManager.swift
```diff
@@ -192,13 +192,37 @@ public struct ContainerManager: Sendable {
         }
     }
 
-    /// Create a new manager with the provided kernel and initfs mount.
+    /// Create a new manager with the provided kernel, initfs mount, image store
+    /// and optional network implementation.
     public init(
         kernel: Kernel,
         initfs: Mount,
+        imageStore: ImageStore,
```
crosbymichael, line 200: It does not look like I have an `init()` that functions as today, just giving me a default.  Do you think this init() should have imagestore as optional, and if it is `nil`, then assigned the default?

crosbymichael (changes requested):

Sources/Containerization/ContainerManager.swift
dcantah: I'm not following the first sentence, but yea we probably could have imagestore as optional, I'm trying to remember why I did it this way

crosbymichael: I would like to have an `init()` like today where I don't have to explicitly specify an ImageStore if I want the default.

dcantah: That is the one below this, albeit it takes in a url.

crosbymichael: If I'm using the default image store, why do I need to know about a URL?

dcantah: I provided two ways of supplying a "store" as the path is used to store more than just images iirc (on phone or I'd check 😂). One being a url and we'll make an imagestore for you based on the url, or the other being to supply an ImageStore yourself if you want. This would allow you to supply the ImageStore's second argument in its constructor if that was helpful.

dcantah: If you just leave the third arg blank (as it is today) the optional url constructor will "win" and we'll just use the default ImageStore.

crosbymichael (approved):

PR #243, opened 2025-08-01
Allow reusing virtiofs mounts

dcantah:
Fwiw, this is a bit odd, but it's also very simple to support and feels odd that this was failing today. This also adds in a tiny bit of logic in hashMountSource to resolve the path of symlinks, so now even symlinks to the same dir can be supplied N times without a new virtiofs device.

### DIFF
--- a/Sources/Containerization/Hash.swift
+++ b/Sources/Containerization/Hash.swift
@@ -18,10 +18,15 @@
 
 import Crypto
 import ContainerizationError
+import Foundation
 
 func hashMountSource(source: String) throws -> String {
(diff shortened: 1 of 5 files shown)


realrajaryan:
Was just looking at apple/container#380. Did we intentionally let this go dormant for some reason?

dcantah:
Just no one reviewed is all

dcantah:
Superseded by https://github.com/apple/containerization/pull/522. The approach in that pr is simpler to me, and I'm lazy to fix the merge confs :)

Issue #246, opened 2025-08-05
[Bug]: A race condition in the `IOPair` class


dkovba:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

The race condition occurs when:

1. Thread A is executing the `relay()` method, actively reading/writing using `self.buffer`.
2. Thread B calls the `close()` method, which calls `self.buffer.deallocate()`.

### Current behavior

Thread A tries to access the deallocated `self.buffer` in thread B, leading to a crash.

### Expected behavior

Access to `self.buffer` should be thread-safe.

### Environment

```markdown
N/A
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #247, opened 2025-08-06
Ensure mutable properties aren't accessed outside their intended thread-safe usage patterns

dkovba:
This PR modifies the access level of certain mutable properties to ensure their thread-safe usage. It doesn't address https://github.com/apple/containerization/issues/246 at this time.


dkovba:
@dcantah Removed a change in `IOPair` to prevent a conflict with https://github.com/apple/containerization/pull/248.

dcantah (approved):

PR #248, opened 2025-08-06
vminitd: Proper pty setup / mount /dev/console

### LINKED ISSUE #145: [Bug]: Need to set up /dev/console

### LINKED ISSUE #246: [Bug]: A race condition in the `IOPair` class

dcantah:
Closes #246 
Closes #145

We need to allocate the pty in the containers ns otherwise you run into funny outcomes later. This change additionally mounts /dev/console by binding the pty opened in the containers namespace onto it. Today /dev/console is solely from devtmpfs which points at /dev/hvc0, or the virtio serial console. This causes systemds output to go to serial/the bootlog
instead of what folks would expect. This should make systemd like us a lot more than it does today.

Another large change that was somewhat required here is the changes in IOPair. EPOLLHUP on a pty is deceptively weird to handle as it could be reopened, so we can't do what we were doing today and just close all the fds. To remedy this I skip exiting on HUP for pty's and instead just cleanup at process exit time, but ensure to drain the pty beforehand.

### DIFF
--- a/vminitd/Sources/vmexec/Console.swift
+++ b/vminitd/Sources/vmexec/Console.swift
@@ -1,0 +1,62 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Musl
+
+class Console {
+    let master: Int32
+    let slavePath: String
+
+    init() throws {
+        let masterFD = open("/dev/ptmx", O_RDWR | O_NOCTTY | O_CLOEXEC)
+        guard masterFD != -1 else {
+            throw App.Errno(stage: "open_ptmx")
+        }
+
+        guard unlockpt(masterFD) == 0 else {
+            throw App.Errno(stage: "unlockpt")
+        }
+
+        guard let slavePath = ptsname(masterFD) else {
+            throw App.Errno(stage: "ptsname")
+        }
+
+        self.master = masterFD
+        self.slavePath = String(cString: slavePath)
+    }
+
+    func configureStdIO() throws {
+        let path = self.slavePath
+        let slaveFD = open(path, O_RDWR)
+        guard slaveFD != -1 else {
+            throw App.Errno(stage: "open_pts")
+        }
+        defer { Musl.close(slaveFD) }
+
+        for fd: Int32 in 0...2 {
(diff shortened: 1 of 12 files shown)


dcantah:
@elijah-wright I can't add you as a reviewer, but I'd still like your feedback on this :)

dcantah:
Will fix the merge conf after that last check in

elijah-wright:
ok I'll look at this right now

vminitd/Sources/vmexec/Console.swift
```diff
@@ -0,0 +1,61 @@
...
+            throw App.Errno(stage: "ptsname")
+        }
+
+        self.master = masterFD
+        self.slavePath = String(cString: slavePath)
+    }
+
+    func configureStdIO() throws {
+        let path = self.slavePath
+        let slaveFD = open(path, O_RDWR)
```
elijah-wright, line 44: is `slaveFD` closed anywhere? I think it's open until `execve`

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -49,92 +49,124 @@ struct RunCommand: ParsableCommand {
...
 
-            // Apply O_CLOEXEC to all file descriptors except stdio.
-            // This ensures that all unwanted fds we may have accidentally
-            // inherited are marked close-on-exec so they stay out of the
-            // container.
-            try App.applyCloseExecOnFDs()
+            try mountConsole(
+                path: pty.slavePath,
+                uid: process.user.uid,
+                gid: process.user.gid
```
elijah-wright: shouldn't this match the default GID of 5 for `tty`?

vminitd/Sources/vmexec/Console.swift
dcantah: No, great catch!

vminitd/Sources/vmexec/RunCommand.swift
dcantah: Another great catch, the chown was a leftover it seems

vminitd/Sources/vmexec/Console.swift
dcantah: We will set it as cloexec shortly below however, but better to just close it right away for clarity.

dkovba (commented):
👍 This should resolve https://github.com/apple/containerization/issues/246.

dcantah:
Going to draft this, as execs seem to have a hang when asking for a pty that isn't seen for the init proc

elijah-wright:
oh I think I was doing some work in that area this afternoon. I'm at work rn but I can look at that in a few hours

dcantah:
Ok, execs fixed. It was just a pipe I wasn't closing.

PR #254, opened 2025-08-07
ContainerizationExtras: Timeout adjustments

dcantah:
Three things:

1. Timeout.run today didn't allow a throwing closure.
2. Add a new constructor for Timeout that takes in a Duration directly so we can be more granular than seconds.
3. Add unit tests because we all love those :)

### DIFF
--- a/Sources/ContainerizationExtras/Timeout.swift
+++ b/Sources/ContainerizationExtras/Timeout.swift
@@ -23,15 +23,38 @@ public struct Timeout {
     /// doesn't finish in the provided `seconds` amount.
     public static func run<T: Sendable>(
         seconds: UInt32,
-        operation: @escaping @Sendable () async -> T
+        operation: @escaping @Sendable () async throws -> T
     ) async throws -> T {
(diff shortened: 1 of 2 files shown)


Sources/ContainerizationExtras/Timeout.swift
```diff
@@ -23,11 +23,11 @@ public struct Timeout {
     /// doesn't finish in the provided `seconds` amount.
     public static func run<T: Sendable>(
         seconds: UInt32,
-        operation: @escaping @Sendable () async -> T
+        operation: @escaping @Sendable () async throws -> T
```
crosbymichael, line 26: Should this be `rethrows` instead of `throws`?

dcantah: I tried this but (it's possible I'm just ignorant) it didn't require me to put throws on the run() call if the closure didn't throw (which is the point of rethrows afair) but this function can always throw a CancellationError depending on how long the closure took to complete, so it was a bit confusing.

crosbymichael (approved):

PR #258, opened 2025-08-09
Handle the same file name in the `fetchBlob` method

dkovba:
With the suggested change, when a file exists, we'll verify whether it has the same content before skipping it.

### DIFF
--- a/Sources/ContainerizationOCI/Client/LocalOCILayoutClient.swift
+++ b/Sources/ContainerizationOCI/Client/LocalOCILayoutClient.swift
@@ -42,7 +42,7 @@ package final class LocalOCILayoutClient: ContentClient {
         }
 
         var hasher = SHA256()
-        let chunkSize = 1024 * 1024
+        let chunkSize = Int(getpagesize()) * 1024
 
         while true {
             let chunk = fileHandle.readData(ofLength: chunkSize)


Sources/ContainerizationOCI/Client/LocalOCILayoutClient.swift
```diff
@@ -53,7 +54,34 @@ package final class LocalOCILayoutClient: ContentClient {
...
                 ])
             }
+        } catch let error as NSError {
+            guard error.code == NSFileWriteFileExistsError else {
+                throw error
+            }
+
+            do {
+                let expectedDigest = try c.digest()
+                let existingData = try Data(contentsOf: file)
```
crosbymichael: We may have issues for large files as this loads the entire file into our memory or is this already an issue with the implementation?

dkovba: We read a file only when we attempt to write data into a file that already exists. This allows to determine whether we should throw an error or treat it as already existing data, which requires this verification.

dkovba: @adityaramani Please confirm that we can't throw an error any time the file already exists.

crosbymichael: If I craft an image with file `/foo` which is 10gb, and then I create another layer that has a file `/foo`, we will hit this branch and cause the code to load the existing 10gb worth of `/foo` into memory?

crosbymichael: You should be able to create an instance of the algo and feed it chunks of data as you read it from the file.

```swift
var sha = SHA256()
sha.update(data: "hi".data(using: .utf8)!)
sha.update(data: "hi".data(using: .utf8)!)

sha.finalize()
```

dkovba: Done

Sources/ContainerizationOCI/Client/LocalOCILayoutClient.swift
```diff
@@ -35,6 +35,26 @@ package final class LocalOCILayoutClient: ContentClient {
...
     }
 
+    private func calculateFileDigest(at url: URL) throws -> SHA256Digest {
+        let fileHandle = try FileHandle(forReadingFrom: url)
+        defer {
+            try? fileHandle.close()
+        }
+
+        var hasher = SHA256()
+        let chunkSize = 1024 * 1024
```
crosbymichael: should we use `getpagesize()` here?

dkovba: 1 MB chunks would be more efficient for larger files than `getpagesize()` (4-16 KB). Or do you suggest something like `Int(getpagesize()) * 1024`?

crosbymichael (approved):

PR #261, opened 2025-08-11
LinuxContainer: Modify UDS mount setup

### LINKED ISSUE #256: [Bug]: Can't mount a host socket when the mount target goes through a symlink.
### I have done the following

### Steps to reproduce

```console
% ls ${SSH_AUTH_SOCK}
/private/tmp/com.apple.launchd.qylZi0PDA7/Listeners

% container run -it --rm --volume "${SSH_AUTH_SOCK}:/var/run/ssh-auth.sock" --env SSH_AUTH_SOCK=/var/run/ssh-auth.sock ubuntu:latest bash
root@3c3908f9-a1de-4ac5-9f36-6f8bf194dab1:/# ls ${SSH_AUTH_SOCK}
ls: cannot access '/var/run/ssh-auth.sock': No such file or directory
root@3c3908f9-a1de-4ac5-9f36-6f8bf194dab1:/#
```

### Current behavior

Mounted socket is inaccessible.

### Expected behavior

```console
% container run -it --rm --volume "$(realpath ${SSH_AUTH_SOCK}):/run/ssh-auth.sock" --env SSH_AUTH_SOCK=/var/run/ssh-auth.sock ubuntu:latest bash 
root@9a626d00-d1ce-4f03-ae17-6784c07dd8b8:/# ls ${SSH_AUTH_SOCK}
/var/run/ssh-auth.sock
```

The this particular case, `ssh-add -l` should work properly, as should `git clone` via ssh.

### Environment

```markdown
- OS: macOS 26 developer preview
- Swift: Apple Swift version 6.2
- Xcode: Xcode 26 beta
```

### Relevant log output

```shell

```

### Code of Conduct

dcantah:
Closes #256

Today the way UDS mounts are done has several problems:

1. They're done before the other container mounts are setup, so it risks them getting shadowed.
2. They're not done in the containers namespaces, so they have a chance to follow a symlink to the VMs rootfs somewhere (/container/rootfs/var/run -> /run in the root of the VM).

This change modifies the logic to setup UDS mounts in a temporary holding place first, and then bind them into place so that they're done at the same time as the other container mounts.

One extra thing this change does is gets rid of the runtime support for UDS mounts. This was originally exposed as I wasn't sure if the container UDS -> host UDS flow would work otherwise, but this works fine today, and we lose the ability to do the runtime mounts in the containers ns'.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -127,6 +127,20 @@ public final class LinuxContainer: Container, Sendable {
         public init() {}
     }
 
+    private struct TemporaryMount {
+        let tempDestination: String
+        let destination: String
+
+        func toBindMount() -> ContainerizationOCI.Mount {
+            .init(
+                type: "bind",
+                source: tempDestination,
+                destination: destination,
+                options: ["bind"]
+            )
+        }
+    }
+
     private let state: Mutex<State>
 
     // Ports to be allocated from for stdio and for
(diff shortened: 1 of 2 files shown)


dcantah:
We still need a "secure join" style mount that resolves symlinks to the containers rootfs and not to their actual absolute path.

PR #265, opened 2025-08-13
vminitd: Add init and execs to cgroup

dcantah:
We currently weren't doing any cgroup setup whatsoever. For the most part this doesn't matter too much, however certain images that may fool around with cgroups don't like this :). Lets do the bare minimum these expect which is to at least have the process running in a nested cg and not apart of the root cg.

### DIFF
--- a/vminitd/Sources/vminitd/CgroupManager.swift
+++ b/vminitd/Sources/vminitd/CgroupManager.swift
@@ -1,0 +1,148 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


vminitd/Sources/vminitd/CgroupManager.swift
```diff
@@ -0,0 +1,148 @@
...
+    case memory
+    case cpuset
+    case cpu
+    case io
+    case hugetlb
+}
+
+// Extremely simple cgroup manager. Our needs are simple for now, and this is
+// reflected in the type.
+internal struct CgroupManager {
```
jglogan, line 33: nit: Cgroupv2Manager?

vminitd/Sources/vminitd/CgroupManager.swift
```diff
@@ -0,0 +1,148 @@
...
+        )
+    }
+
+    private static func writeValue(path: URL, value: String, fileName: String) throws {
+        let file = path.appending(path: fileName)
+        let fd = open(file.path, O_WRONLY, 0)
+        if fd == -1 {
+            throw Error.errno(errno: errno, message: "failed to open \(file.path)")
+        }
+        defer { Musl.close(fd) }
```
jglogan, line 74: n00b question: why `open()` but `Musl.close()`?

PR #270, opened 2025-08-19
Cgroup2Manager: Various adjustments

dcantah:
Systemd loves to move the root cgroups processes to one it created named /init.scope and then write the root cgs subtree_control file. Because of this we can't just add exec processes to the cg we made for the container anymore as we'll get EBUSY. We should follow where the init processes cg is actually at (/proc/pid/cgroup) and add it there.


jglogan (approved):

dcantah:
@afbjorklund Forgot to tag you as this one may be of interest. I believe this gets systemd in order finally. We'll make a tag and trial in container

PR #271, opened 2025-08-22
Handle warnings as errors

dkovba:
To handle warnings as errors when building, use `make`. To allow warnings when building, use `make allow-warnings`.
To handle warnings as errors when testing, use `make test`. To allow warnings when testing, use `make WARNINGS_AS_ERRORS=false test`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@
 # Build configuration variables
 # The default version ID 0.0.0 indicates a local development build or PRB
 BUILD_CONFIGURATION ?= debug
+WARNINGS_AS_ERRORS ?= true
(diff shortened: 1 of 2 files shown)


crosbymichael:
Instead of an additional make target, can we just unify all the targets with:

- `make WARNINGS_AS_ERRORS=false` 
- `make WARNINGS_AS_ERRORS=false test`
- `make WARNINGS_AS_ERRORS=false init`
- - `make WARNINGS_AS_ERRORS=false ...`

What do you think?

dkovba:
I do not mind, but this would make building with warnings allowed less convenient. @dcantah Do you have a preference?

dcantah:
Michael's idea is fine with me. It'll hopefully be rare enough (meaning the need to build with warnings) that it won't matter for ergonomics much.

PR #285, opened 2025-09-10
Add IO utilities for traversing dirs and reading files from EXT4 devices

wlan0:
Read/Write directly from EXT4 block device

### DIFF
--- a/Sources/ContainerizationEXT4/EXT4Reader+IO.swift
+++ b/Sources/ContainerizationEXT4/EXT4Reader+IO.swift
@@ -1,0 +1,515 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


Sources/ContainerizationEXT4/EXT4Reader+IO.swift
```diff
@@ -0,0 +1,515 @@
...
+        return
+            children
+            .map { $0.0 }
+            .filter { $0 != "." && $0 != ".." }
+            .sorted()
+    }
+
+    /// Read bytes from a regular file at `path` starting at `offset`.
+    /// If `count` is nil, reads to EOF. Returns exactly the requested bytes (or less at EOF).
+    public func readFile(
```
dcantah, line 110: Can we try and adopt a non-filehandle like model where the data is returned to the caller (or at least have a method where the user passes in a buffer for us as well)?

wlan0: yeah passing in a buffer is a good idea!

wlan0: @dcantah updated the PR!

PR #288, opened 2025-09-11
feat: add proxy utility

mayooot:


jglogan:
@mayooot Sorry, I still am not seeing a verified signature on the commit?

jglogan:
See: https://docs.github.com/en/authentication/troubleshooting-commit-signature-verification

Here's a bit from ChatGPT on gotchas:

#### Common gotchas

- The user.email in your Git config still needs to match one of your GitHub account emails (or your no-reply GitHub email).
- If you have multiple SSH keys, specify the right one in `user.signingkey`. You can check which is loaded with:
`ssh-add -l`

PR #294, opened 2025-09-17
Add filesystem notification (FSNotify) support

realrajaryan:
Addresses apple/container#141, where containers don't receive filesystem events on mounted volumes, preventing incremental rebuilds and other file-watching features. This PR implements the guest-side components for FSNotify. Host-side implementation in the container repo will complete the pipeline.

Summary:
- Add gRPC protocol definitions for filesystem event notifications
- Implement guest-side event handler that generates authentic Linux inotify events
- Add CLI testing tool (`cctl fsnotify`) and integration test infrastructure


Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -60,6 +60,8 @@ service SandboxContext {
   rpc Sync(SyncRequest) returns (SyncResponse);
   // Send a signal to a process via the PID.
   rpc Kill(KillRequest) returns (KillResponse);
+  // Notify guest of filesystem events from host.
+  rpc NotifyFileSystemEvent(NotifyFileSystemEventRequest) returns (NotifyFileSystemEventResponse);
```
jglogan: Wondering if we should have messages for FileSystemEvent and FileSystemEventList, and have the request type deal with lists for performance. We'd need to work out how do deal with errors... perhaps a list of error responses, where each contains a FileSystemEvent, and an error string. An empty response list means no errors.

Sources/Integration/VMTests.swift
```diff
@@ -305,6 +307,94 @@ extension IntegrationSuite {
...
+        try await container.start()
+
+        // Get the vminitd agent to send notifications
+        let connection = try await container.dialVsock(port: 1024)  // Default vminitd port
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        let agent = Vminitd(connection: connection, group: group)
+
+        // Test 1: CREATE event
+        print("Testing CREATE event...")
+        let createResponse = try await agent.notifyFileSystemEvent(path: "/mnt/test/new_file.txt", eventType: .create)
```
jglogan: Should probably test create event on an existing file.

The reasoning is that from the point of view of your macOS filesystem monitor it will:
- Get an event from macOS that a file was created.
- At this point the file exists and should be accessible to the container (not sure what the latency for this is)
- Trigger a create event for the already-existing file.

vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -921,6 +921,72 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
...
+
+    private func generateSyntheticInotifyEvent(
+        path: String,
+        eventType: Com_Apple_Containerization_Sandbox_V3_FileSystemEventType
+    ) async throws {
+        switch eventType {
+        case .modify:
+            // Touch file to update timestamp -> generates IN_ATTRIB event
+            let now = Date()
+            try FileManager.default.setAttributes(
```
jglogan: We shouldn't change the modification date. The modification date was already changed, we just want to notify.

vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -921,6 +921,72 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
...
+                [.modificationDate: now],
+                ofItemAtPath: path
+            )
+
+        case .create:
+            // Use chmod with same permissions to generate IN_ATTRIB event
```
jglogan: Better to error out if we can't get permissions.

chmodding a file from 400 to 644 wouldn't be ideal, nor would changing a file from 755 to 644.

Sources/Containerization/SandboxContext/SandboxContext.proto
james-lawrence: or change it to a stream?

dcantah: A stream is likely the right answer.

dcantah:
@realrajaryan Can you take a peek at the CI fail, it's for the new test added

Sources/Integration/VMTests.swift
```diff
@@ -305,6 +307,94 @@ extension IntegrationSuite {
...
+        }
+
+        // Test 2: MODIFY event on existing file
+        print("Testing MODIFY event...")
+        let modifyResponse = try await agent.notifyFileSystemEvent(path: "/mnt/test/existing.txt", eventType: .modify)
+        guard modifyResponse.success else {
+            throw IntegrationError.assert(msg: "MODIFY event failed: \(modifyResponse.error)")
+        }
+
+        // Verify file timestamp was updated
```
jglogan: Are we really checking file timestamps here and in the previous test? It looks like we just check existence by running stat but don't actually do anything to see that the event was processed.

jglogan: the could run a container like:

```
container run -it --rm -v "${PWD}/notify-data:/mnt/notify-data" alpine:latest sh -c "apk add inotify-tools ; inotifywait -m /mnt/notify-data -e modify --format '%e %f'"
```

Or there are one-liners in python or node that could call `inotify()` in a loop and display the events.

jglogan: tests seem to be failing as well

Sources/Integration/VMTests.swift
```diff
@@ -305,6 +308,106 @@ extension IntegrationSuite {
...
+            // Install inotify-tools and monitor the mount point for events
+            config.arguments = [
+                "/bin/sh", "-c",
+                """
+                apk add --no-cache inotify-tools > /dev/null 2>&1 && \
+                timeout 2 inotifywait -m /mnt -e modify,create,delete --format '%e %f' 2>/dev/null &
+                INOTIFY_PID=$!
+                sleep 0.1
+                # Trigger a modify event that should be detected
+                touch /mnt/test-inotify.txt
```
jglogan: This isn't useful except to test whether local changes work. It doesn't test the agent path.

Sources/Integration/VMTests.swift
```diff
@@ -305,6 +308,106 @@ extension IntegrationSuite {
...
+                """,
+            ]
+            config.stdout = inotifyBuffer
+        }
+
+        try await inotifyProcess.start()
+
+        // While inotify is running, send FSNotify events that should trigger inotify
+        try await Task.sleep(for: .milliseconds(200))
+        let _ = try await agent.notifyFileSystemEvent(path: "\(vmMountPath)/test-inotify.txt", eventType: .modify)
```
jglogan, line 372: I would expect that we'd start the inotify process, then run tests 1 and 2, and expect to see two modify events on `existing.txt`.

Sources/Integration/VMTests.swift
```diff
@@ -305,6 +308,106 @@ extension IntegrationSuite {
...
+
+        let _ = try await inotifyProcess.wait()
+        let inotifyOutput = String(data: inotifyBuffer.data, encoding: .utf8) ?? ""
+
+        // Verify that inotify detected the modify event
```
jglogan: But is the event from L360, or L372? Are there two events or one in the output?

jglogan (changes requested):

Sources/Integration/VMTests.swift
jglogan: Try commenting out 360-361 and the test fails

Sources/Integration/VMTests.swift
```diff
@@ -305,6 +308,106 @@ extension IntegrationSuite {
...
+        // Get the vminitd agent to send notifications
+        let connection = try await container.dialVsock(port: 1024)  // Default vminitd port
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        let agent = Vminitd(connection: connection, group: group)
+
+        // Calculate the hashed tag name for the mount source and mount in VM
+        let mountTag = try hashMountSource(source: directory.path)
+        let vmMountPath = "/tmp/fsnotify-test"
+
+        try await agent.mount(.init(type: "virtiofs", source: mountTag, destination: vmMountPath))
```
jglogan: This gets the test to pass but isn't correct.

Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -282,3 +284,23 @@ message KillRequest {
...
 
 message KillResponse { int32 result = 1; }
+
+enum FileSystemEventType {
+  CREATE = 0;
+  DELETE = 1;
+  LINK = 2;
+  UNLINK = 3;
+  MODIFY = 4;
+  UNDEFINED = 99;
```
jglogan: What does an UNDEFINED event represent? When would such an event be propagated through the gRPC API?

Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -282,3 +284,23 @@ message KillRequest {
...
+}
+
+message NotifyFileSystemEventRequest {
+  string path = 1;
+  FileSystemEventType event_type = 2;
+  string container_id = 3;
```
jglogan, line 369: It's not clear why, but it seems we use snake case for just about everything in this file, except a container ID is `containerID` 🤷

Sources/Containerization/Vminitd.swift
```diff
@@ -367,6 +367,32 @@ extension Vminitd {
         try await Task.sleep(for: .milliseconds(10))
         try await self.sync()
     }
+
+    /// Send a filesystem event notification to the guest.
+    public func notifyFileSystemEvent(
```
jglogan, line 459: Since we're using streaming, the API here should be some sort of collection of Com_Apple_Containerization_Sandbox_V3_NotifyFileSystemEventRequest (or a typealias of it).

Sources/Integration/Suite.swift
```diff
@@ -123,7 +137,8 @@ struct IntegrationSuite: AsyncParsableCommand {
         let platform = Platform(arch: "arm64", os: "linux", variant: "v8")
 
         let fs: Containerization.Mount = try await {
-            let fsPath = Self.testDir.appending(component: "rootfs.ext4")
+            let imageHash = Self.hashImageReference(reference)
```
jglogan: Just use `image.digest` in place of manually hashing the image reference.

Sources/Integration/VMTests.swift
```diff
@@ -348,6 +351,105 @@ extension IntegrationSuite {
...
+            path: "/mnt/existing.txt",
+            eventType: .modify,
+            containerID: id
+        )
+        guard modifyResponse.success else {
+            throw IntegrationError.assert(msg: "MODIFY event failed: \(modifyResponse.error)")
+        }
+        print("MODIFY event succeeded")
+
+        // Wait for events to be processed
```
jglogan: Remove all the debug prints before merging.

Sources/cctl/FSNotifyCommand.swift
```diff
@@ -0,0 +1,125 @@
...
+
+import ArgumentParser
+import Containerization
+import ContainerizationError
+import Foundation
+import GRPC
+import NIOCore
+import NIOPosix
+
+extension Application {
```
jglogan, line 25: Does this command work? What commands should I use to test it locally?

Sources/Integration/VMTests.swift
```diff
@@ -348,6 +351,105 @@ extension IntegrationSuite {
...
+
+        // Clean up
+        try await agent.close()
+        try await group.shutdownGracefully()
+        try await container.stop()
+
+        print("All FSNotify events tested successfully")
+    }
+
+    private func createFSNotifyTestDirectory() throws -> URL {
```
jglogan: Do we really need this, or could we instead use createMountDirectory since we're only working with one file? If we do need to create a subdirectory we could do it as part of the test.

Sources/Integration/VMTests.swift
```diff
@@ -348,6 +351,105 @@ extension IntegrationSuite {
...
+        print("Waiting for inotify events to be detected...")
+        try await Task.sleep(for: .seconds(1))
+
+        let inotifyOutput = String(data: inotifyBuffer.data, encoding: .utf8) ?? ""
+        print("=== Final Container Output ===")
+        print(inotifyOutput)
+        print("=== End Container Output ===")
+
+        // Verify that inotify detected the modify event
+        guard inotifyOutput.contains("change") && inotifyOutput.contains("existing.txt") else {
```
jglogan: It seems that this should below the delete event, and rather than checking for the presence of any `change` and any `existing.txt`, we should be ensuring that the actual output exactly matches the expected output.

Sources/Integration/Suite.swift
dcantah: Good idea, I'd suggested doing this verbally but I forgot we have the image object already at this point.

dcantah: What's good about just using the digest is in the (no idea why we'd do this but) odd scenario where it's the same image, just a different reference/tag, we still wouldn't unpack for no reason if we already had.

vminitd/Sources/vminitd/ManagedContainer.swift
```diff
@@ -28,6 +36,370 @@ actor ManagedContainer {
...
+                guard written == buffer.count else {
+                    throw ContainerizationError(.internalError, message: "Failed to write response to parent")
+                }
+            }
+        }
+
+        private func readResponseFromChild(socket: Int32) throws -> (UInt32, UInt8)? {
+            // Read event_id:4
+            var eventID: UInt32 = 0
+            guard read(socket, &eventID, 4) == 4 else { return nil }
```
dcantah: I don't see a mechanism to get notified of when we have events to read so it seems we're just doing blocking read(2)'s on a swift concurrency thread. For low core count guests this will likely cause us to hang 😅. You're not supposed to block any of the swift concurrency threads, as they're always supposed to be making forward progress or hitting some suspension point. We could use nio for these socket reads and that'd at least get us into shape

vminitd/Sources/vminitd/ManagedContainer.swift
```diff
@@ -28,6 +36,370 @@ actor ManagedContainer {
     private let log: Logger
     private let bundle: ContainerizationOCI.Bundle
     private var execs: [String: ManagedProcess] = [:]
+    private var namespaceWorker: NamespaceWorker?
+
+    /// Worker child process that runs in container's namespace for filesystem operations
+    private final class NamespaceWorker: @unchecked Sendable {
```
jglogan: A few things here:

- Is there a reason for nesting this class? It's adding about 350 lines to ManagedContainer, which makes it harder for someone unfamiliar with the code to reason about either. I can move it outside of ManagedContainer and still compile it, so let's move it into its own file.
- I could see this being called NamespaceWorker if it facilitated any kind of work taking place between the guest agent and namespace, which could be a useful abstraction later on. For now let's change the name to describe specifically the work being done.
- We need to manage the concurrency between guest agent, worker parent, and worker child such that `@unchecked` isn't needed for Sendable. In particular, we can't do blocking socket `read()` calls in Task context.

vminitd/Sources/vminitd/ManagedContainer.swift
```diff
@@ -28,6 +36,370 @@ actor ManagedContainer {
...
+                    // Child failed to send signal
+                    close(parentSocket)
+                    self.parentSocket = nil
+                    var status: Int32 = 0
+                    waitpid(pid, &status, 0)
+                    self.childPID = nil
+                    throw ContainerizationError(.internalError, message: "Child process failed to start")
+                }
+
+                if signal == 0xFF {
```
jglogan: Extract constants for these and then you can get rid of the comments.

Also, let's not call it a "signal" as that has a specific meaning that is not this.

vminitd/Sources/vminitd/ManagedContainer.swift
```diff
@@ -28,6 +36,370 @@ actor ManagedContainer {
...
+                    close(parentSocket)
+                    self.parentSocket = nil
+                    var status: Int32 = 0
+                    waitpid(pid, &status, 0)
+                    self.childPID = nil
+                    throw ContainerizationError(.internalError, message: "Child process sent unexpected signal: \(signal)")
+                }
+
+                // Start response reader task
+                self.responseReaderTask = Task { [weak self] in
```
jglogan: This is no different from:

```swift
            self.responseReaderTask = Task {
                await self.readChildResponses()
            }
```

jglogan (changes requested):
Let's get the concurrency working so that it's not `@unchecked`. See Danny's suggestion about using NIO for the socket communication.

vminitd/Sources/vminitd/FilesystemEventWorker.swift
```diff
@@ -0,0 +1,378 @@
...
+import NIOPosix
+import Synchronization
+
+#if canImport(Musl)
+import Musl
+#elseif canImport(Glibc)
+import Glibc
+#endif
+
+final class FilesystemEventWorker: @unchecked Sendable {
```
dcantah: We should try and make this actually sendable. Just shove all the mutable state in a Mutex (we can reuse the one we have for pendingEvents, but just make a struct with all of this data on it, and call the field state or something similar.

vminitd/Sources/vminitd/FilesystemEventWorker.swift
```diff
@@ -0,0 +1,378 @@
...
+
+        var sockets: [Int32] = [0, 0]
+        guard socketpair(AF_UNIX, SOCK_STREAM, 0, &sockets) == 0 else {
+            throw ContainerizationError(.internalError, message: "Failed to create socketpair: errno \(errno)")
+        }
+
+        let parentSocket = sockets[0]
+        let childSocket = sockets[1]
+
+        let pid = fork()
```
dcantah: I'm skeptical this will work alright. Here's a decent article on some of the pitfalls https://thorstenball.com/blog/2014/10/13/why-threads-cant-fork/

dcantah: We'd ideally want to exec a new process that we pass the other half of the socket pair to.

dcantah: (the new process does not physically need to be a new binary, we can just edit vminitd to have a fs-notify subcommand and fork+exec ourselves)

Sources/Containerization/Vminitd.swift
```diff
@@ -428,6 +430,46 @@ extension Vminitd {
...
+        guard results.count == events.count else {
+            throw ContainerizationError(.internalError, message: "Expected \(events.count) responses, got \(results.count)")
+        }
+
+        return results
+    }
+
+    public func notifyFileSystemEvent(
+        path: String,
+        eventType: Com_Apple_Containerization_Sandbox_V3_FileSystemEventType,
```
jglogan: nit, would a type alias for this make sense as well?

Sources/Integration/ContainerTests.swift
```diff
@@ -969,6 +971,87 @@ extension IntegrationSuite {
...
+        }
+
+        try await Task.sleep(for: .seconds(1))
+
+        let inotifyOutput = String(data: inotifyBuffer.data, encoding: .utf8) ?? ""
+
+        let expectedLines = ["change hi.txt", "change hi.txt"]
+        let actualLines = inotifyOutput.trimmingCharacters(in: .whitespacesAndNewlines).components(separatedBy: .newlines).filter { !$0.isEmpty }
+
+        guard actualLines.count >= expectedLines.count else {
```
jglogan: Is there a reason why the test `expectedLines == actualLines` doesn't work reliably?

Sources/Integration/Suite.swift
```diff
@@ -151,16 +151,17 @@ struct IntegrationSuite: AsyncParsableCommand {
     var kernel: String = "./bin/vmlinux"
 
     @Option(name: .shortAndLong, help: "Maximum number of concurrent tests")
-    var maxConcurrency: Int = 8
+    var maxConcurrency: Int = 6
```
jglogan: Why this change?

vminitd/Sources/vminitd/FilesystemEventWorker.swift
```diff
@@ -0,0 +1,293 @@
...
+        guard state.withLock({ $0.childPID }) == nil else {
+            throw ContainerizationError(.invalidState, message: "FilesystemEventWorker already started")
+        }
+
+        var sockets: [Int32] = [0, 0]
+        guard socketpair(AF_UNIX, SOCK_STREAM, 0, &sockets) == 0 else {
+            throw ContainerizationError(.internalError, message: "Failed to create socketpair: errno \(errno)")
+        }
+
+        let parentSocket = sockets[0]
```
jglogan, line 68: We're not defer closing for any of this because of exec and/or we're manually closing some of these at L105-106?

vminitd/Sources/vminitd/ManagedContainer.swift
```diff
@@ -15,11 +15,20 @@
...
 
 import Cgroup
+import Containerization
 import ContainerizationError
 import ContainerizationOCI
 import ContainerizationOS
 import Foundation
 import Logging
+import NIOCore
+import Synchronization
```
jglogan: What is Synchronization used for in this file?

vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -1078,6 +1079,57 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
             $0.result = r
         }
     }
+
+    func notifyFileSystemEvent(
+        requestStream: GRPCAsyncRequestStream<Com_Apple_Containerization_Sandbox_V3_NotifyFileSystemEventRequest>,
+        responseStream: GRPCAsyncResponseStreamWriter<Com_Apple_Containerization_Sandbox_V3_NotifyFileSystemEventResponse>,
+        context: GRPC.GRPCAsyncServerCallContext
+    ) async throws {
+        for try await request in requestStream {
```
jglogan, line 1091: Maybe an optimization for later - read the stream into an array and send the entire batch to the worker?

Might be best to look at the end-to-end path at that point and work out what's most efficient.

Sources/Integration/Suite.swift
dcantah: I was going to make it in another pr, I can get that in before this so it's not part of the diff.

vminitd/Sources/vminitd/FilesystemEventWorker.swift
```diff
@@ -0,0 +1,293 @@
...
+final class FilesystemEventWorker: Sendable {
+    private static let handshakeReady: UInt8 = 0xAA
+    private static let handshakeFailure: UInt8 = 0xFF
+
+    private let containerID: String
+    private let containerPID: Int32
+    private let eventLoop: EventLoop
+    private let shouldStop: Atomic<Bool> = Atomic(false)
+
+    // Cross-thread state (accessed from any thread)
```
jglogan: accessed from any parent thread (that is, any `vminitd init` thread), right?

vminitd/Sources/vminitd/FilesystemEventWorker.swift
```diff
@@ -0,0 +1,293 @@
...
+        let eventTypeValue = UInt32(eventType.rawValue)
+
+        var buffer = Data()
+        buffer.append(contentsOf: withUnsafeBytes(of: eventTypeValue.bigEndian) { Data($0) })
+        buffer.append(contentsOf: withUnsafeBytes(of: pathLen.bigEndian) { Data($0) })
+        buffer.append(pathData)
+        buffer.append(contentsOf: withUnsafeBytes(of: eventID.bigEndian) { Data($0) })
+
+        try buffer.withUnsafeBytes { bytes in
+            let written = write(socket, bytes.bindMemory(to: UInt8.self).baseAddress, buffer.count)
```
jglogan: So for inbound stuff it looks like we're using NIO but for outbound data we're doing direct socket writes? Is is possible to set up the outbound pipeline and use it?

vminitd/Sources/vminitd/Server+GRPC.swift
jglogan: (doesn't block merge)

abdshn:
I see that this MR is reverted back and also issue is resolved, is there a new issue created ?

realrajaryan:
> I see that this MR is reverted back and also issue is resolved, is there a new issue created ?

Issue is still open, we’re working on it :)

PR #295, opened 2025-09-17
examples: add ctr-example

egernst:
Add very basic example of running a Linux container using Containerization


katiewasnothere:
Should we add something in the makefile that grabs the vmlinux file? And add info on how to run the example in the readme?

egernst:
Yes, and probably the initfs as well - it’s current left to the user since I use reference for image store.

dcantah:
`make fmt`

dcantah:
Still says formatting is off :(

PR #301, opened 2025-09-25
Kernel: Fix `make`

### LINKED ISSUE #296: [Bug]: cannot get `make` to succeed in the `kernel/` folder
### I have done the following

### Steps to reproduce

Preamble: I ended up here because my high-level goal is to build my own kernel with Landlock enabled that works with `container run --kernel` as discussed on https://github.com/apple/container/issues/527. I thought I could possibly enable a few options in `config-arm64` and use this to build.

---

I am on `51ef9f81fef574bbd815d4f5560157297b0a4067` with no local changes.

I upgraded to macOS Tahoe and ran through all of the build instructions on https://github.com/apple/containerization/blob/main/README.md. The output of `ls -l bin` in the root of my repo is as follows after following those instructions:

```
total 1646264
-rw-------@ 1 mbolin  staff    1271252 Sep 19 14:23 boot.log
-rwxr-xr-x@ 1 mbolin  staff   52371200 Sep 19 14:22 cctl
-rwxr-xr-x@ 1 mbolin  staff   52151040 Sep 19 14:22 containerization-integration
-rw-r--r--@ 1 mbolin  staff  536870912 Sep 19 14:23 init.block
-rw-r--r--@ 1 mbolin  staff  172513280 Sep 19 14:22 init.rootfs.tar.gz
-rw-r--r--@ 1 mbolin  staff   14750208 Sep 19 14:15 vmlinux
```

### Current behavior

As instructed on [`kernel/README.md`](https://github.com/apple/containerization/blob/51ef9f81fef574bbd815d4f5560157297b0a4067/kernel/README.md), I am trying to run `make` in the `kernel/` folder, but it fails as follows:

```
$ make
../bin/cctl run \
		--cpus 8 \
		--memory 16384 \
		--fs-size 32768 \
		--kernel ../bin/vmlinux \
		--init ../bin/init.block \
		--image ghcr.io/apple/test-images/kernel-build:latest \
		--mount .:/kernel \
		--cwd /kernel \
		-- /bin/bash -c "./build.sh"
Error: internalError: "Failed to exec [--init ../bin/init.block --image ghcr.io/apple/test-images/kernel-build:latest --mount .:/kernel --cwd /kernel -- /bin/bash -c ./build.sh] Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory""
```

It fails with `No such file or directory`, but I don't know what file it is referring to.

I am not sure if this is the problem, but I am unclear on whether this is something I should be able to access:

https://github.com/apple/containerization/blob/51ef9f81fef574bbd815d4f5560157297b0a4067/kernel/Dockerfile#L1

Certainly if I run `container build .` in the `kernel/` folder, I get:

```
[+] Building 0.0s (0/1)
 => [resolver] fetching image...ghcr.io/apple/backpack-test/kernel-build:latest                                                 0.0s
Error: internalError: "HTTP request to https://ghcr.io/v2/apple/backpack-test/kernel-build/manifests/latest failed with response: 401 Unauthorized. Reason: Access denied or wrong credentials. No credentials found for host ghcr.io"
```

### Expected behavior

I would expect `make` to succeed.

### Environment

```markdown
- OS: Tahoe 26.0
- Swift: swift-driver version: 1.127.14.1 Apple Swift version 6.2 (swiftlang-6.2.0.19.9 clang-1700.3.19.1)
- Xcode: Version 26.0 (17A324)
```

### Relevant log output

```shell

```

### Code of Conduct

dcantah:
Fixes #296

Today the kernel subdirectory is somewhat useless. The dockerfile we had tried to use an image that isn't actually on ghcr, so everything would kind of just crap out. This change swaps things so that we build a local image that has all the bits needed to compile the kernel, and then uses that image to build the resulting kernel and plop it in the current working directory. This change does add a reliance on `container` being installed however, but for folks that need to build a custom kernel this doesn't seem like too much of an issue to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,10 @@ Containerization depends on specific versions of `grpc-swift` and `swift-protobu
 make protos
 ```
 
+## Building a kernel
+
+If you'd like to build your own kernel please see the instructions in the [kernel directory](./kernel/README.md).
+
 ## Documentation
 
 Generate the API documentation for local viewing with:
--- a/kernel/Dockerfile
+++ b/kernel/Dockerfile
@@ -1,9 +1,0 @@
-FROM ghcr.io/apple/backpack-test/kernel-build:latest as build
-COPY . /src
-WORKDIR /src
-RUN make prepare-kernel
-RUN make build-kernel
-
-FROM scratch
-LABEL org.opencontainers.image.source=https://github.com/apple/containerization
-COPY --from=build /src/kernel.arm64 /
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -13,19 +13,29 @@
 # limitations under the License.
 
 KSOURCE ?= https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.14.9.tar.xz
-KIMAGE ?= ghcr.io/apple/test-images/kernel-build:latest
+KIMAGE ?= kernel-build:0.1
+CURDIR := $(shell pwd)
 
-all:
+.DEFAULT_GOAL := all
+
+.PHONY: all
+all: kernel-build-image
+all: kernel-build
+
+.PHONY: kernel-build-image
+kernel-build-image:
+	container build image/ -f image/Dockerfile -t ${KIMAGE}
+
+.PHONY: kernel-build
+kernel-build:
 ifeq (,$(wildcard source.tar.xz))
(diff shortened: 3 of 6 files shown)


jglogan (approved):
I can follow the docs and build a kernel, as advertised.

PR #311, opened 2025-10-04
Add configurable concurrent layer downloads

sbhavani:
Adds `maxConcurrentDownloads` parameter to `ImageStore.pull()`.

Currently hardcoded to 8 concurrent downloads. This makes it configurable with a default of 3.

Performance improvement: ~3-5x faster pulls for multi-layer images with higher concurrency.

### DIFF
--- a/Sources/Containerization/Image/ImageStore/ImageStore+Import.swift
+++ b/Sources/Containerization/Image/ImageStore/ImageStore+Import.swift
@@ -30,13 +30,15 @@ extension ImageStore {
         let contentStore: ContentStore
         let progress: ProgressHandler?
         let name: String
+        let maxConcurrentDownloads: Int
 
-        public init(name: String, contentStore: ContentStore, client: ContentClient, ingestDir: URL, progress: ProgressHandler? = nil) {
+        public init(name: String, contentStore: ContentStore, client: ContentClient, ingestDir: URL, progress: ProgressHandler? = nil, maxConcurrentDownloads: Int = 3) {
             self.client = client
             self.ingestDir = ingestDir
             self.contentStore = contentStore
             self.progress = progress
             self.name = name
+            self.maxConcurrentDownloads = maxConcurrentDownloads
         }
 
         /// Pull the required image layers for the provided descriptor and platform(s) into the given directory using the provided client. Returns a descriptor to the Index manifest.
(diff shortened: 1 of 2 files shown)


dcantah:
>Performance improvement: ~3-5x faster pulls for multi-layer images with higher concurrency.

Can you provide some details on the images you tried and actual time deltas?

sbhavani:
@dcantah I haven't tested many images but here's a real world example with more modest speedups:

 Image: nvcr.io/nvidia/cuda:12.6.2-base-ubuntu24.04
 Size: 866.9 MB

  | Concurrency    | Time (real) | Speedup vs Sequential          |
  |----------------|-------------|--------------------------------|
  | 1 (sequential) | 21.674s     | 1.0x baseline                  |
  | 3 (default)    | 17.761s     | 1.22x faster (18% improvement) |
  | 6              | 16.514s     | 1.31x faster (24% improvement) |

  Time Savings:
  - Concurrency 3: Saves ~3.9 seconds (18% faster)
  - Concurrency 6: Saves ~5.2 seconds (24% faster)
  
System info:
Model Name: MacBook Air
Chip: Apple M3
Memory: 16 GB

Network info:
- Download: 297.5 Mbps
- Upload: 5.8 Mbps

sbhavani:
Sorry I overlooked this before but the default was set at 8 which is pretty reasonable for my current connection over VPN (se below). 

  - Concurrency 8: 10.267s
  - Concurrency 12: 9.239s  (10% improvement)
  
But there may be cases where users have slower connection and need lower concurrency or faster connection and can handle higher concurrency values. Also helps to align with Docker which sets a default of 3.

dcantah:
Can you run `make fmt` and repush?

dcantah:
cc @adityaramani, I'd like your eyes on this. This seems decent to support, it's a nice little toggle.

dcantah:
@adityaramani Was the 8 chosen originally based on some heuristic we had? Don't have the history

dcantah:
Oop, @sbhavani didn't notice this but we require signed commits (gpg, ssh etc.). Can you sign with one of those methods and repush? I'd just squash down to one commit and sign that.

dcantah:
@sbhavani It seems the squashed commit still isn't signed :( I don't mean a "signed off by" line if it wasn't clear, but an actual signature via ssh/gpg etc.

PR #314, opened 2025-10-06
Fixes proxy access for image registry types.

### LINKED ISSUE #255: [Bug]: image fetch honors `HTTP_PROXY` but not other standard variables.
### I have done the following

### Steps to reproduce

All of `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`, both upper and lower case, should be checked. This should be in a utility function so that clients in container can use common code.

See: 

- related issue apple/container#466 with expected and actual results
- [a pretty decent reference article](https://about.gitlab.com/blog/we-need-to-talk-no-proxy/)
- [relevant code](https://github.com/apple/containerization/blob/main/Sources/ContainerizationOCI/Client/RegistryClient.swift#L113)

### Current behavior

see apple/container#466

### Expected behavior

see apple/container#466

### Environment

```markdown
- OS: 15.6
- Swift: 16.4
- Xcode: container CLI version 0.3.0 (build: release, commit: 3fcf647)
```

### Relevant log output

```shell

```

### Code of Conduct

jglogan:
- Closes #255.
- The async-http-client library proxy configuration is all-or-nothing; HTTPClient doesn't support selecting the proxy based individual request URLs.
- Creates an HTTPClientProvider protocol that provides an HTTPClient given a URL and exposes shutdown functions for cleaning up owned HTTPClients.
- Implements a ProxyAwareHTTPClientProvider that creates HTTPClients based on the standard http_proxy/https_proxy environment variables, and provides the appropriate client based on URL scheme and host exclusion matching as determined by the no_proxy environment variable.
- Updates RegistryClient, ImageStore, and ContainerManager to use HTTPClientProvider instead on creating a fixed-configuration HTTPClient with incorrect proxy support in RegistryClient.
- Updates `cctl` and tests to use the proxy client provider.
- Removes ProxyUtils class as this didn't match up with what was needed for the HTTPClient proxy configuration.


Sources/cctl/cctl.swift
```diff
@@ -70,6 +75,19 @@ struct Application: AsyncParsableCommand {
...
 }
 
+struct CommandContext {
+    let appRoot: URL
+    let eventLoopGroup: any EventLoopGroup
+    let httpClientProvider: HTTPClientProvider
+    let contentStore: ContentStore
+    let imageStore: ImageStore
+
+    func shutdown() throws {
```
crosbymichael: This approach seems very error prone, even for instances where I'm not pulling an image. It will be very easy to forget to call shutdown.

crosbymichael (changes requested):
With all the changes it feels pretty leaky all over the codebase.  Maybe we can chat on this one real-time and figure something out. I'll spend some cycles on it today.

Sources/cctl/cctl.swift
jglogan: Per our discussion, simplified to keep RegistryClient opinionated and just fix the proxy handling there and in ProxyUtils.

Sources/ContainerizationOCI/Client/RegistryClient.swift
```diff
@@ -109,17 +111,13 @@ public final class RegistryClient: ContentClient {
...
-                return nil
-            }
-            return .server(host: host, port: port)
-        }()
-        httpConfiguration.proxy = proxyConfig
+
+        // proxy configuration assumes all client requests will go to `base` URL
+        self.proxyURL = ProxyUtils.proxyFromEnvironment(scheme: scheme, host: host)
+        if let proxyURL = self.proxyURL, let proxyHost = proxyURL.host {
+            httpConfiguration.proxy = HTTPClient.Configuration.Proxy.server(host: proxyHost, port: proxyURL.port ?? 8080)
```
crosbymichael: is 8080 correct here or should it be 80?

jglogan: That should be 80 given that I changed other things to be go-like. I'll rebase and fix.

Originally that's what I thought curl used and I was doing curl-style semantics. But 8080 is wrong there as well – curl actually defaults to 1080 🤷 (though the docs aren't clear whether this applies to just socks proxies)

jglogan: 443 for https scheme, 80 otherwise now

PR #318, opened 2025-10-07
Add support for container statistics

### LINKED ISSUE #138: [Request]: Container Stats Functionality

dcantah:
Closes #138

This rounds up cgroup and network stats (via netlink) and exposes them via an agent rpc and on LinuxContainer. While this is not an entirely accurate view into the full resources being used by the container given the virtualized nature (doesn't account for vcpu, device etc overhead) it should give an accurate overview of the workload resource usage in the guest.

This change besides the stated goal also removes the interfaceStatistics rpc and just moves these network stats into the new containerStatistics one. Related to that, it also renames the InterfaceStatistics struct to NetworkStatistics and moves this into ContainerStatistics as a nested struct.


Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -298,3 +286,69 @@ message KillRequest {
...
+  repeated string container_ids = 1;  // Empty = all containers
+}
+
+message ContainerStatisticsResponse {
+  repeated ContainerStats containers = 1;
+}
+
+message ContainerStats {
+  string container_id = 1;
+  ProcessStats processes = 2;
```
crosbymichael: `s/processes/process` ?

Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -298,3 +286,69 @@ message KillRequest {
 }
 
 message KillResponse { int32 result = 1; }
+
+message ContainerStatisticsRequest {
+  repeated string container_ids = 1;  // Empty = all containers
```
crosbymichael, line 291: Do you think a one_of here would be better?

Sources/Containerization/ContainerStatistics.swift
```diff
@@ -0,0 +1,177 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// Statistics for a container.
+public struct ContainerStatistics: Sendable {
+    public var id: String
+    public var processes: ProcessStatistics
```
crosbymichael: s//process/

Sources/Containerization/SandboxContext/SandboxContext.proto
dcantah: I don't think this is too bad, I think if someone called this with no ids provided it'd be odd for us to do nothing. Perhaps this should be named filter though?

dcantah: I left this as is for now, we can always change though

crosbymichael (approved):

PR #323, opened 2025-10-09
Handle EOF on one side of a UnixSocketRelay.

jglogan:
- Current implementation shuts down everything as soon as EOF is detected on DispatchSourceRead for the relay. This is problematic for, say an HTTP request where the client makes a request, and calls `shutdown(fd, SHUT_WR)` to close the send side, but it expects to be able to keep calling `recv()` to get the response.
- Changed cancel handlers so that the one that sees that both sources have been cancelled then closes both the UDS and vsock fds.
- Updated vminitd VsockProxy to also do `shutdown(fd, SHUT_WR)` for read hangup or EOF, and only close the underlying fds and unwire pollers on full hangup, broken pipe, error, or when both sides half-close.

### DIFF
--- a/Sources/Containerization/UnixSocketRelay.swift
+++ b/Sources/Containerization/UnixSocketRelay.swift
@@ -391,7 +391,16 @@ extension SocketRelay {
                         "dstFd": "\(destinationFd)",
                     ])
                 source.cancel()
-                shutdown(destinationFd, SHUT_WR)
+                if shutdown(destinationFd, SHUT_WR) != 0 {
+                    log?.info(
+                        "failed to shut down reads",
+                        metadata: [
+                            "errno": "\(errno)",
+                            "sourceFd": "\(sourceFd)",
+                            "dstFd": "\(destinationFd)",
+                        ]
+                    )
+                }
             }
             return
         }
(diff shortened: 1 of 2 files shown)


Sources/Containerization/UnixSocketRelay.swift
```diff
@@ -159,6 +159,10 @@ extension SocketRelay {
         let hostSocket = try Socket(type: socketType)
         try hostSocket.listen()
 
+        log?.info("listening on host UDS", metadata: [
```
jglogan: Need to remove log messages and/or move to debug before moving out of draft.

jglogan:
Converting back to draft. I'm seeing that with this change things like this hang:

```bash
echo "hello" | docker exec -i 80fcb781f6a0 cp /dev/stdin /tmp/hello.txt
```

vminitd/Sources/vminitd/VsockProxy.swift
```diff
@@ -191,57 +235,79 @@ extension VsockProxy {
...
 
                     if mask.isHangup {
+                        eofFromClient = true
+                        eofFromServer = true
+                    } else if mask.isRhangup && !eofFromClient {
+                        // half close, shut down client to server transfer
+                        // we should see no more EPOLLIN events on the client fd
+                        // and no more EPOLLOUT events on the server fd
+                        eofFromClient = true
+                        shutdown(serverFile.fileDescriptor, SHUT_WR)
```
dcantah: Check if failed and log something (as we can't really bubble anything up from here).

PR #330, opened 2025-10-14
Socket: Add fd receiving

dcantah:
Via scm_rights. Useful for supporting spawning runc as an OCI runtime as that's how the pty is passed to the client.

### DIFF
--- a/Sources/CShim/socket_helpers.c
+++ b/Sources/CShim/socket_helpers.c
@@ -1,0 +1,33 @@
+/*
+ * Copyright © 2025 Apple Inc. and the Containerization project authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "socket_helpers.h"
+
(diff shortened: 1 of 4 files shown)


Sources/CShim/socket_helpers.c
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "socket_helpers.h"
+
+struct cmsghdr* CZ_CMSG_FIRSTHDR(struct msghdr *msg) {
```
dcantah, line 19: Only reason is you can't call c macros in swift. So they need to be wrapped in a function we actually can.

Tests/ContainerizationOSTests/SocketTests.swift
```diff
@@ -0,0 +1,121 @@
...
+        let receivedFileHandle = try recvSocket.receiveFileDescriptor()
+        defer { try? receivedFileHandle.close() }
+
+        #expect(receivedFileHandle.fileDescriptor != originalFD, "Received FD should be different")
+        #expect(receivedFileHandle.fileDescriptor >= 0, "Received FD should be valid")
+
+        let data = try receivedFileHandle.readToEnd()
+        #expect(data != nil, "Should be able to read from received FD")
+
+        let receivedContent = String(data: data!, encoding: .utf8)
```
katiewasnothere, line 118: Just a heads up that a failed `#expect` does not end the test. If the above check that `data != nil` fails, we will still get to this line, which I'm assuming may panic since data is nil.

dcantah: Changed to #require!

katiewasnothere (approved):

PR #333, opened 2025-10-15
Add runc process launch wrapper

dcantah:
Somewhat simple and missing events and some other bits I believe, but it's a start. Maybe should use the async subprocess type as well instead of our homegrown Command.

### DIFF
--- a/vminitd/Sources/vminitd/Runc/Runc.swift
+++ b/vminitd/Sources/vminitd/Runc/Runc.swift
@@ -1,0 +1,756 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


dcantah:
Will fix ci in the morning, just forgot formatting

vminitd/Sources/vminitd/Runc/Runc.swift
```diff
@@ -0,0 +1,744 @@
...
+        cmd.stderr = stderr ?? outPipe.fileHandleForWriting
+
+        // FIXME: pdeathSignal handling if Command supported it.
+
+        if setpgid {
+            cmd.attrs.setPGroup = true
+        }
+
+        try cmd.start()
+        let exitStatus = try cmd.wait()
```
crosbymichael: We will need a way to have `wait` be pluggable. A lot of the production use cases the caller will either be a sub-reaper or running as pid 1 in the `vminitd` case.  This will cause the reaper and this wait to race between who gets the exist signal first.

dcantah: Shit you're right, need to hook into our reaper

crosbymichael (approved):

PR #341, opened 2025-10-20
VirtualMachineManager: Rework protocol

dcantah:
Today this protocols create method is just odd. Our only implementation of it immediately casts to LinuxContainer, failing if it cannot do so. I think what might make more sense is to pass in a configuration itself with core parameters that we expect every vmm to be able to support, and then in a specific implementation they can continue to cast this type to a specific one to possibly extract some extra configuration values (rosetta for VZ for example).

This rework will also make it simpler to support a Pod type, as the vm setup is identical and simple.

### DIFF
--- a/Sources/Containerization/VMConfiguration.swift
+++ b/Sources/Containerization/VMConfiguration.swift
@@ -1,0 +1,68 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


Sources/Containerization/VMConfiguration.swift
```diff
@@ -0,0 +1,67 @@
...
+    public var cpus: Int
+    /// The memory in bytes to allocate.
+    public var memoryInBytes: UInt64
+    /// The network interfaces to attach.
+    public var interfaces: [any Interface]
+    /// Mounts organized by metadata ID (e.g. container ID).
+    /// Each ID maps to an array of mounts for that workload.
+    public var mountsByID: [String: [Mount]]
+    /// Whether nested virtualization should be enabled.
+    public var nestedVirtualization: Bool
```
crosbymichael: Should this be treated the same as rosetta support?

dcantah: Yea, was debating on this. I can't think of many vmms that don't support this now, but to be safe probably.

dcantah: I'll move

dcantah: I decided to keep this. If the VMM doesn't support it, I've documented it that the VMM MUST return .unsupported.

PR #343, opened 2025-10-22
Introduce new experimental LinuxPod type

### LINKED ISSUE #319: [Request]: Explore pod-like type

While the 1-1 container-vm LinuxContainer type we have is the main feature of this library, it might be nice in some scenarios to support > 1 container in the guest if they're all related workloads (or need to share a pid ns). This would probably be an experimental API to start.

### Code of Conduct

dcantah:
Closes #319

Introduces a new type capable of running > 1 container in the guest. The API mostly follows LinuxContainer, and each individual container can be addressed via any of the methods that require you to pass the containerID as the first param. Today there's no support for namespace sharing, but that shouldn't be terrible to support.

### DIFF
--- a/Sources/Containerization/LinuxPod.swift
+++ b/Sources/Containerization/LinuxPod.swift
@@ -1,0 +1,736 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+#if os(macOS)
+import ContainerizationError
+import ContainerizationExtras
+import ContainerizationOCI
+import Foundation
+import Logging
+import Synchronization
+
+import struct ContainerizationOS.Terminal
+
+/// NOTE: Experimental API
+///
+/// `LinuxPod` allows managing multiple Linux containers within a single
+/// virtual machine. Each container has its own rootfs and process, but
+/// shares the VM's resources (CPU, memory, network).
+public final class LinuxPod: Sendable {
+    /// The identifier of the pod.
+    public let id: String
+
+    /// Configuration for the pod.
+    public let config: Configuration
+
+    /// The configuration for the LinuxPod.
+    public struct Configuration: Sendable {
+        /// The amount of cpus for the pod's VM.
+        public var cpus: Int = 4
+        /// The memory in bytes to give to the pod's VM.
+        public var memoryInBytes: UInt64 = 1024.mib()
+        /// The network interfaces for the pod.
+        public var interfaces: [any Interface] = []
+        /// The DNS configuration for the pod.
+        public var dns: DNS?
+        /// Whether nested virtualization should be turned on for the pod.
+        public var virtualization: Bool = false
+        /// Optional file path to store serial boot logs.
+        public var bootlog: URL?
+
+        public init() {}
+    }
+
+    /// Configuration for a container within the pod.
+    public struct ContainerConfiguration: Sendable {
+        /// Configuration for the init process of the container.
+        public var process = LinuxProcessConfiguration()
+        /// Optional per-container CPU limit (can exceed pod total for oversubscription).
+        public var cpus: Int?
+        /// Optional per-container memory limit in bytes (can exceed pod total for oversubscription).
+        public var memoryInBytes: UInt64?
+        /// The hostname for the container.
+        public var hostname: String = ""
+        /// The system control options for the container.
+        public var sysctl: [String: String] = [:]
+        /// The mounts for the container.
+        public var mounts: [Mount] = LinuxContainer.defaultMounts()
+        /// The Unix domain socket relays to setup for the container.
+        public var sockets: [UnixSocketConfiguration] = []
+
+        public init() {}
+    }
+
+    private struct PodContainer: Sendable {
+        let id: String
+        let rootfs: Mount
+        let config: ContainerConfiguration
+        var state: ContainerState
+        var process: LinuxProcess?
+
+        enum ContainerState: Sendable {
+            case registered
+            case created
+            case started
+            case stopped
+            case errored
+        }
+    }
+
+    private let state: AsyncMutex<State>
+
+    // Ports to be allocated from for stdio and for
+    // unix socket relays that are sharing a guest
+    // uds to the host.
+    private let hostVsockPorts: Atomic<UInt32>
+    // Ports we request the guest to allocate for unix socket relays from
+    // the host.
+    private let guestVsockPorts: Atomic<UInt32>
+
+    private struct State: Sendable {
+        var phase: Phase
+        var containers: [String: PodContainer]
+    }
+
+    private enum Phase: Sendable {
+        /// The pod has been created but no live resources are running.
+        case initialized
+        /// The pod's virtual machine has been setup and the runtime environment has been configured.
+        case created(CreatedState)
+        /// An error occurred during the lifetime of this class.
+        case errored(Swift.Error)
+
+        struct CreatedState: Sendable {
+            let vm: any VirtualMachineInstance
+            let relayManager: UnixSocketRelayManager
+        }
+
+        func createdState(_ operation: String) throws -> CreatedState {
+            switch self {
+            case .created(let state):
+                return state
+            case .errored(let err):
+                throw err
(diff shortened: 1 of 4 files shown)


Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+    public init(
+        _ id: String,
+        vmm: VirtualMachineManager,
+        logger: Logger? = nil,
+        configuration: (inout Configuration) throws -> Void
+    ) throws {
+        self.id = id
+        self.vmm = vmm
+        self.hostVsockPorts = Atomic<UInt32>(0x1000_0000)
+        self.guestVsockPorts = Atomic<UInt32>(0x1000_0000)
```
realrajaryan, line 169: `guestVsockPorts` isn't used anywhere right now, do we need to keep this?

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+                // Check if the vm is even still running
+                if createdState.vm.state == .stopped {
+                    container.state = .stopped
+                    state.containers[containerID] = container
+                    return
+                }
+
+                try await createdState.vm.withAgent { agent in
+                    // Kill all processes in the container
+                    try await agent.kill(pid: -1, signal: SIGKILL)
```
realrajaryan: Wouldn't this nuke processes in _all_  containers in the pod? I know we do this in `LinuxContainer` but we have only one container per vm there. 

`Cgroup2Manager.swift` has a `kill()` method we could use instead ig, as we already set per container cgroupsPath here.

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+                for id in state.containers.keys {
+                    state.containers[id]?.state = .created
+                }
+
+                state.phase = .created(.init(vm: vm, relayManager: relayManager))
+            } catch {
+                try? await vm.stop()
+                state.phase.setErrored(error: error)
+                throw error
+            }
```
realrajaryan, line 367: If `create()` fails after relays are started, we only call `vm.stop()` but never `relayManager.stopAll()`. This can leave host-side relays running even though the VM is gone. We should call `relayManager.stopAll()` in the catch block before stopping the VM, same as we do in `stop()`.

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+                                )
+                            }
+                        }
+
+                        try? await process.delete()
+                        container.process = nil
+                        container.state = .stopped
+                        state.containers[containerID] = container
+                    }
+                }
```
realrajaryan, line 523: We use `try?` while killing/waiting/umounting and still mark the container `.stopped`, but if any step fails, state drifts from reality (proc may still run, mounts/relays may remain).

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+        } else {
+            socket.source = rootInGuest.appending(path: socket.source.path)
+        }
+
+        let port = self.hostVsockPorts.wrappingAdd(1, ordering: .relaxed).oldValue
+        try await relayManager.start(port: port, socket: socket)
+        try await relayAgent.relaySocket(port: port, configuration: socket)
+    }
+
+    private static func setupIO(
```
realrajaryan: nit: this is pretty much same as `LinuxContainer.setupIO`, let's make it a shared helper. (maybe there are more such cases, should prob look)

Sources/Integration/PodTests.swift
realrajaryan, line 1: We should also probably add a test to verify that containers are actually isolated from each other.

Sources/Containerization/LinuxPod.swift
dcantah: I used LinuxContainer as a base, and upon further inspection, I don't even use guestVsockPorts over there either 🤣. I'll fix LinuxContainer in a separate change, but will mirror it here.

Sources/Containerization/LinuxPod.swift
dcantah: Agree

Sources/Containerization/LinuxPod.swift
dcantah: Yea this is a blunder/crummy copy paste from pod level stop

Sources/Containerization/LinuxPod.swift
dcantah: Good idea

Sources/Containerization/LinuxPod.swift
realrajaryan: nit: we could also eventually move port allocation into `UnixSocketRelayManager` with a small reserve/release model so ports freed by stopped containers can be reused.

rn it’s monotonic (`wrappingAdd`), which isn’t a problem in practice but does mean long-lived pods keep incrementing the counter.

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+        /// Whether nested virtualization should be turned on for the pod.
+        public var virtualization: Bool = false
+
+        public init() {}
+    }
+
+    /// Configuration for a container within the pod.
+    public struct ContainerConfiguration: Sendable {
+        /// Configuration for the init process of the container.
+        public var process = LinuxProcessConfiguration.init()
```
realrajaryan: nit: `init()` is redundant

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+    public var memoryInBytes: UInt64 {
+        config.memoryInBytes
+    }
+
+    /// Network interfaces of the pod.
+    public var interfaces: [any Interface] {
+        config.interfaces
+    }
+
+    /// Add a container to the pod. This must be called before `create()`.
```
realrajaryan, line 243: nit: maybe worth enforcing the documented behavior (`guard case .initialized = state.phase`) to fail early if called after `create()`, rather than relying on callers to follow the doc comment

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+            }
+            try await process.closeStdin()
+        }
+    }
+
+    /// Relay a unix socket for a container.
+    public func relayUnixSocket(_ containerID: String, socket: UnixSocketConfiguration) async throws {
+        let createdState = try await self.state.withLock { state in
+            try state.phase.createdState("relayUnixSocket")
+        }
```
realrajaryan: `relayUnixSocket()` doesn't check if containerID actually exists in the pod. If you pass a nonexistent/invalid containerID, it will just construct the path `/run/container/some-dogslop-id/rootfs` and then will fail with a confusing filesystem error when trying to setup the relay. Should prob validate the container exists first so we can fail fast with a clear error

Sources/Containerization/LinuxPod.swift
```diff
@@ -0,0 +1,761 @@
...
+            spec.linux?.resources?.cpu = LinuxCPU(
+                quota: Int64(cpus * 100_000),
+                period: 100_000
+            )
+        }
+        if let memoryInBytes = config.memoryInBytes {
+            spec.linux?.resources?.memory = LinuxMemory(
+                limit: Int64(memoryInBytes)
+            )
+        }
```
realrajaryan: Setting `cpus = 0` or `memoryInBytes = 0` will write `cpu.max = 0 / memory.max = 0` in cgroups, starving the container. Should validate > 0 or treat 0 as no limit or unset, wdyt?

dcantah: Good catch.. previously virt.framework would yell at us prior to the cg toggles so we didn't need our own validation

Sources/Containerization/LinuxPod.swift
dcantah: Fair

Sources/Containerization/LinuxPod.swift
dcantah: We can treat this as unset/no limit for now

Sources/Containerization/LinuxPod.swift
dcantah: Right now we don't support adding containers at runtime so it's also not too big of an issue. But agree, my original plan was to have the ports be managed by that type itself. There's a comment somewhere in this project (or maybe it was a commit msg) that hints at it. As usual, it was put on the backburner 😆

Sources/Containerization/LinuxPod.swift
dcantah: It's mostly because below all of these "ignore this error" dance, the vm itself is actually stopped `try await createdState.vm.stop()`, which if that succeeds all of the container state is irrelevant and back to square one as we setup

dcantah:
Fixed almost everything locally (reason for the resolves) will push in the AM after fixing a new hang from the changes 🐸

crosbymichael (approved):

Sources/Containerization/LinuxPod.swift
dcantah: I left this one as technically there's not really an issue, just a funny mishap. I will fix this and LinuxContainers in the same change

realrajaryan (approved):

PR #347, opened 2025-10-24
Add custom decoder inits for various oci types that use omitempty

katiewasnothere:
Many fields on the various OCI types use "omitempty" for encoding and decoding the json representation in golang. This PR adds custom json decoder functions to allow for behavior similar to "omitempty".

### DIFF
--- a/Tests/ContainerizationOCITests/OCISpecTests.swift
+++ b/Tests/ContainerizationOCITests/OCISpecTests.swift
@@ -1,0 +1,146 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Testing
+
+@testable import ContainerizationOCI
+
+struct OCISpecTests {
+    @Test func minimalSpecDecode() throws {
+        let version = "1.2.3"
+        let minSpec =
+            """
+                {
+                    "ociVersion": "\(version)"
+                }
+            """
+
+        guard let data = minSpec.data(using: .utf8) else {
+            Issue.record("test spec is not valid: \(minSpec)")
+            return
+        }
+
+        let decodedSpec = try JSONDecoder().decode(ContainerizationOCI.Spec.self, from: data)
+        #expect(decodedSpec.version == version)
+        #expect(decodedSpec.hooks == nil)
+        #expect(decodedSpec.process == nil)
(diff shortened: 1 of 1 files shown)


Sources/ContainerizationOCI/Spec.swift
```diff
@@ -61,6 +61,25 @@ public struct Spec: Codable, Sendable {
...
+        self.version = try container.decode(String.self, forKey: .version)
+        self.hooks = try container.decodeIfPresent(Hook.self, forKey: .hooks)
+        self.process = try container.decodeIfPresent(Process.self, forKey: .process)
+        if let hostname = try container.decodeIfPresent(String.self, forKey: .hostname) {
+            self.hostname = hostname
+        }
+        if let domainname = try container.decodeIfPresent(String.self, forKey: .domainname) {
+            self.domainname = domainname
+        }
+        self.mounts = try container.decode([Mount].self, forKey: .mounts)
```
realrajaryan: Should use `decodeIfPresent` instead of `decode`. See the [Go spec](https://github.com/opencontainers/runtime-spec/blob/main/specs-go/config.go), `Mounts` has `json:"mounts,omitempty"`, meaning it’s optional. This decoder will throw when `mounts` is missing from the JSON, but the OCI spec allows minimal configs without any mounts. You can safely default to an empty array:

```swift
self.mounts = try container.decodeIfPresent([Mount].self, forKey: .mounts) ?? []
```

Sources/ContainerizationOCI/Spec.swift
```diff
@@ -209,6 +321,16 @@ public struct Mount: Codable, Sendable {
         self.uidMappings = uidMappings
         self.gidMappings = gidMappings
     }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.type = try container.decode(String.self, forKey: .type)
+        self.source = try container.decode(String.self, forKey: .source)
+        self.destination = try container.decode(String.self, forKey: .destination)
+        self.options = try container.decode([String].self, forKey: .options)
```
realrajaryan: Similar to what I mentioned above, `Mount` itself should also use `decodeIfPresent` for `type`, `source`, and `options`. In the [Go spec](https://github.com/opencontainers/runtime-spec/blob/main/specs-go/config.go), all three are tagged with `omitempty`, so they can be missing. Using `decode` here makes them required and will throw on minimal configs.

Also bc the struct initializer doesn’t provide defaults for `type` and `source`, it’s worth adding them to align with the spec. Without this some valid configs (like bind or tmpfs mounts that omit type/source) will prob fail to decode

Sources/ContainerizationOCI/Spec.swift
katiewasnothere: Updated

Sources/ContainerizationOCI/Spec.swift
katiewasnothere: Updated

katiewasnothere:
> I wonder if we should add a test for decoding a minimal OCI spec JSON (eg. one without mounts) to catch issues like this. The existing OCI tests don't cover runtime spec decoding

@realrajaryan I'm not sure how useful tests would be for these changes since the tests would just be agreeing with the code. But I can add some as a backup check if you'd like?

dcantah:
@katiewasnothere I'd think the test(s) (decode minimal spec) prior to these changes would throw an error however. It would be nice to have

katiewasnothere:
@realrajaryan @dcantah Tests added. LMK if you'd prefer any additional tests.

PR #350, opened 2025-10-24
Throw an error in `StandardIO.resize()`

dkovba:
Throw an error in `StandardIO.resize()`, which is unsupported.


dcantah:
Doesn't compile right now

dkovba:
> Doesn't compile right now

Fixed!

PR #354, opened 2025-10-24
Rename `process` to `command`

dkovba:
Rename `process` to `command` (assuming we agree on the change).


dcantah:
I'd prefer process. I understand the name of the type we're using to instantiate it is command, but that's just a decision we made likely to mirror Go's exec.Cmd. The running entity after starting the command is a process however, and I think it reads nicer everywhere but the initial creation point.

dcantah (approved):
Team prefers command

PR #368, opened 2025-10-28
Make Index.mediaType optional to comply with OCI spec

blark:
The `mediaType` field in the `Index` struct was defined as a required field, but according to the [OCI Image Index Specification](https://github.com/opencontainers/image-spec/blob/main/image-index.md), this field is optional.

This caused failures when loading OCI archives where the `index.json` omits the top-level `mediaType` field, which is valid per the spec. Tools like skopeo can generate such archives.

## Error before fix
```
keyNotFound(CodingKeys(stringValue: "mediaType", intValue: nil))
```

## Changes
- Changed `Index.mediaType` from `String` to `String?`
- Updated initializer to accept optional `mediaType` parameter
- Added comment documenting that field is optional per OCI spec

## Testing
Verified that OCI archives without a top-level `mediaType` field in `index.json` now load successfully.

Fixes https://github.com/apple/container/issues/330

### DIFF
--- a/Sources/ContainerizationOCI/Index.swift
+++ b/Sources/ContainerizationOCI/Index.swift
@@ -25,6 +25,7 @@ public struct Index: Codable, Sendable {
     public let schemaVersion: Int
 
     /// mediaType specifies the type of this document data structure e.g. `application/vnd.oci.image.index.v1+json`
+    /// This field is optional per the OCI Image Index Specification (omitempty)
     public let mediaType: String
 
     /// manifests references platform specific manifests.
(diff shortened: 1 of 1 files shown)


katiewasnothere:
Hi @blark thanks for the contribution! Could you follow a similar pattern to what was done in this PR https://github.com/apple/containerization/pull/347 for Index.mediaType?

We're hoping to only use optionals in the spec definitions for pointer types. Other types that are defined omitempty in the oci spec, we will rely on the use of custom JSON decoders to ensure similar behavior to omitempty for non-pointer types in go.

blark:
Thanks for the feedback @katiewasnothere! I've updated the PR to follow the pattern from #347:

- Added a custom `init(from decoder: Decoder)` implementation
- Kept `mediaType` as `String` (not optional)
- Uses `decodeIfPresent` with `""" as the default (matching the string pattern in #347)
- Maintains the same behavior for handling omitted fields per OCI spec

PR #374, opened 2025-10-29
Integration: Add UDS test

dcantah:
We didn't have one 🐸. A much better test eventually would be to actually relay something through the socket, but we need something in the image that can actually do this. We can probably write a small statically linked bin that can do this later.


realrajaryan (approved):
we should track the request to add more or better ones later on like you suggested

PR #377, opened 2025-10-31
LinuxContainer: Add new constructor

dcantah:
Add a constructor that allows just passing in a configuration without doing the closure dance.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -64,6 +64,34 @@ public final class LinuxContainer: Container, Sendable {
         public var bootlog: URL?
 
         public init() {}
+
+        public init(
+            process: LinuxProcessConfiguration,
+            cpus: Int = 4,
+            memoryInBytes: UInt64 = 1024.mib(),
+            hostname: String = "",
+            sysctl: [String: String] = [:],
+            interfaces: [any Interface] = [],
+            sockets: [UnixSocketConfiguration] = [],
+            mounts: [Mount] = LinuxContainer.defaultMounts(),
+            dns: DNS? = nil,
+            hosts: Hosts? = nil,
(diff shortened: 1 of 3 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -214,6 +246,32 @@ public final class LinuxContainer: Container, Sendable {
...
+    ///   - vmm: The virtual machine manager that will handle launching the VM for the container.
+    ///   - configuration: The container configuration specifying process, resources, networking, and other settings.
+    ///   - logger: Optional logger for container operations.
+    public init(
+        _ id: String,
+        rootfs: Mount,
+        vmm: VirtualMachineManager,
+        configuration: LinuxContainer.Configuration,
+        logger: Logger? = nil
+    ) throws {
```
realrajaryan: This is marked `throws` but contains no throwing code, do we need it?

Sources/Integration/ContainerTests.swift
```diff
@@ -987,6 +987,31 @@ extension IntegrationSuite {
         }
     }
 
+    func testDifferentConstructor() async throws {
```
realrajaryan: nit: `testConfigurationObjectConstructor`/`testNonClosureConstructor` would be a better name, lol

Sources/Containerization/LinuxContainer.swift
dcantah: Nope, we can get rid of it good catch.

Sources/Integration/ContainerTests.swift
dcantah: LOL Fair fair

dcantah: This name doesn't really help if we introduce a third constructor either 😆

PR #379, opened 2025-11-02
Move to grpc-swift V2

### LINKED ISSUE #51: Use grpc-swift v2
There's a newer grpc-swift variant that touts higher perf and a slicker API. I have a local change for this that I can tidy up.

dcantah:
- Closes #51
- Closes #517 as this applies similar changes. 
- No API changes which is nice. The largest change code wise is in Vminitd. The connection loop is setup immediately in the new API by calling `runConnections()` on the client, and this blocks until either the task it's running on is cancelled or `beginGracefulShutdown` is called. Because of this, either we'd have to modify our API to have VirtualMachineAgent implementations call some `run()` method, or we just put `runConnections()` in an internal task on Vminitd. I went for the latter approach as it's simpler.


dkovba:
@dcantah We can proceed with this PR.

jglogan (approved):

PR #381, opened 2025-11-04
Improvements to the process and network management commands 

dkovba:
Improves the process and network management commands.

### DIFF
--- a/vminitd/Sources/vminitd/Server+GRPC.swift
+++ b/vminitd/Sources/vminitd/Server+GRPC.swift
@@ -47,20 +47,33 @@ private let _sync = Glibc.sync
 extension ContainerizationError {
     func toGRPCStatus(operation: String) -> GRPCStatus {
         let message = "\(operation): \(self)"
-        switch self.code {
-        case .invalidArgument:
-            return GRPCStatus(code: .invalidArgument, message: message)
-        case .notFound:
-            return GRPCStatus(code: .notFound, message: message)
-        case .exists:
-            return GRPCStatus(code: .alreadyExists, message: message)
-        case .cancelled:
-            return GRPCStatus(code: .cancelled, message: message)
-        case .unsupported:
-            return GRPCStatus(code: .unimplemented, message: message)
-        default:
-            return GRPCStatus(code: .internalError, message: message)
-        }
+        let code: GRPCStatus.Code = {
+            switch self.code {
+            case .invalidArgument:
+                return .invalidArgument
+            case .notFound:
+                return .notFound
+            case .exists:
+                return .alreadyExists
+            case .cancelled:
+                return .cancelled
+            case .unsupported:
+                return .unimplemented
+            case .unknown:
+                return .unknown
+            case .internalError:
+                return .internalError
+            case .interrupted:
+                return .unavailable
+            case .invalidState:
+                return .failedPrecondition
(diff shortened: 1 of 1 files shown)


vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -525,17 +525,45 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
...
 
-        if !request.hasContainerID {
-            throw ContainerizationError(
-                .invalidArgument,
-                message: "processes in the root of the vm not implemented"
-            )
-        }
+        do {
+            if !request.hasContainerID {
+                throw ContainerizationError(
```
crosbymichael, line 569: Why did you move this inside of the `do` as your `catch` will change this invalid arg to an internal error?

dkovba: Fixed.

vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -488,17 +488,30 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
...
+                "createProcess",
+                metadata: [
+                    "id": "\(request.id)",
+                    "containerID": "\(request.containerID)",
+                    "error": "\(err)",
+                ])
+            switch err.code {
+            case .invalidArgument:
+                throw GRPCStatus(code: .invalidArgument, message: "createProcess: failed to create process: \(err)")
+            default:
```
crosbymichael: Why are we getting lower fidelity on these error codes? Since we are catching a `ContainerizationError` it's going to have the original code already as that actually happened. Throwing a new error causes us to lose a bit of information.  Is there a reason to do this through this PR?

dkovba: We are catching `ContainerizationError` to convert it to `GRPCStatus`. Without this conversion, all errors would become generic GRPC internal errors.

dkovba: It seems reasonable to implement `toGRPCStatus` in `ContainerizationError` to reduce the amount of code for error conversion and to convert all possible errors.

crosbymichael: Oh I see now.  Many of the codes should map well from cz error to grpc status.  You could do something like:

```swift
extension ContainerizationError {
    /// Convert the ContainerizationError into a GRPCStatus result.
    var status: GRPCStatus {
        let code: GRPCStatus.Code = {
            switch self.code {
            case .notFound: .notFound
            case .exists: .alreadyExists
            case .unsupported: .unimplemented
            case .unknown: .unknown
            case .internalError: .internalError
            case .interrupted: .unavailable
            case .invalidState: .failedPrecondition
            case .invalidArgument: .invalidArgument
            case .timeout: .deadlineExceeded
            default: .internalError
            }
        }()
        return GRPCStatus(
            code: code,
            message: self.description,
            cause: self
        )
    }
}
```

dkovba: This approach loses the operation context compared to what I've already implemented. However, we can handle more error codes.

jglogan (approved):

PR #383, opened 2025-11-05
Set the default `PATH` and fix the bug in retrieving the `PATH` value

dkovba:
- We're already setting the default `HOME` and `TERM`. Setting the default `PATH` as well seems reasonable.
- This PR also fixes the bug in retrieving the `PATH` value (`PATH=` has to be removed, and `=` should be allowed in folder names).


vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -1141,6 +1141,10 @@ extension Initd {
             seenSuppGids.insert($0).inserted
         }
 
+        if !process.env.contains(where: { $0.hasPrefix("PATH=") }) {
+            process.env.append("PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
+        }
```
katiewasnothere, line 1146: A default path is already set on the process if no path env variable is given to the request [here](https://github.com/apple/containerization/blob/bec1008f4de1e416ea0e112b57b2357875ec7d25/vminitd/Sources/vmexec/vmexec.swift#L81), but I do agree that we should be setting a default in a more obvious place. 

Maybe instead of hardcoding the value here though, we reference the variable Danny added like is done [here](https://github.com/apple/containerization/blob/bec1008f4de1e416ea0e112b57b2357875ec7d25/Sources/Containerization/Vminitd.swift#L55) so there's a centralized definition of the default path.

dkovba: Replaced the old way of setting the default with the new one in this PR, reusing the existing constant.

dcantah: The setenv(PATH you removed is for vminitd itself, not a container. Please add it back

dcantah: (In standardSetup)

dkovba: I've been still testing the change. Reverted.

dcantah (approved):

PR #402, opened 2025-11-12
feat(registry): Add custom ca certificate override

sebimarkgraf:
Closely related to https://github.com/apple/container/issues/305 I would like to override the used SSL TrustRoots via standard env variables.

This here would add this configuration and would give an entrypoint for an implementation of 305 to provide CLI flags or similar.

This has no tests yet, as this would require setting up something like a MITM proxy when testing against a registry.
As I am unfamiliar with the codebase, I would be willing to do this, but would require a first nudge on where to best implement this.

To actually use this, we would need to add the allowed env variables to the `container system start` command env filter.

### DIFF
--- a/Sources/ContainerizationOCI/Client/RegistryClient.swift
+++ b/Sources/ContainerizationOCI/Client/RegistryClient.swift
@@ -22,6 +22,7 @@ import Foundation
 import Logging
 import NIO
 import NIOHTTP1
+import NIOSSL
 
 #if os(macOS)
 import Network
@@ -66,7 +67,8 @@ public final class RegistryClient: ContentClient {
         reference: String,
         insecure: Bool = false,
         auth: Authentication? = nil,
-        logger: Logger? = nil
+        tlsConfiguration: TLSConfiguration? = nil,
+        logger: Logger? = nil,
     ) throws {
         let ref = try Reference.parse(reference)
         guard let domain = ref.resolvedDomain else {
@@ -86,7 +88,8 @@ public final class RegistryClient: ContentClient {
             scheme: scheme,
             port: port,
             authentication: auth,
-            retryOptions: Self.defaultRetryOptions
+            retryOptions: Self.defaultRetryOptions,
+            tlsConfiguration: tlsConfiguration,
         )
     }
 
@@ -98,7 +101,8 @@ public final class RegistryClient: ContentClient {
         clientID: String? = nil,
         retryOptions: RetryOptions? = nil,
         bufferSize: Int = Int(4.mib()),
-        logger: Logger? = nil
+        tlsConfiguration: TLSConfiguration? = nil,
+        logger: Logger? = nil,
     ) {
         var components = URLComponents()
         components.scheme = scheme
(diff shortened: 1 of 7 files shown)


Sources/ContainerizationOCI/Client/RegistryClient.swift
```diff
@@ -44,6 +45,23 @@ public struct RetryOptions: Sendable {
     }
 }
 
+func makeEnvironmentAwareTLSConfiguration() -> TLSConfiguration {
```
dcantah: I wonder if `RegistryClient`'s constructor should just take in a `TLSConfiguration` and `container` just copies this logic you have here and supplies the `TLSConfiguration` we crafted from the envvars. An alternative would be to take in an enum that the cases could be `.fromEnvironment`, `.custom(TLSConfiguration)` and so on. I'm leaning towards the first (adding tlsConfiguration as a parameter) as it's the most flexible

sebimarkgraf: Probably sensible to have the first approach:
Otherwise this enum would at some point repeat most of the options already covered in TLSConfiguration itself.
Also this could include a `--skip-ca-verification` cli flag that just allows skipping as a temporary workaround.

Also, if I move this up one level it could be easier to extract and merge it somewhere upstream in a more refined way.

sebimarkgraf: But, this would mean that both `container` in `LoginCommand` and `containerization` in `ImageStore` need to instantiate a TLSConfig and hand it to the `RegistryClient`.

sebimarkgraf: Reworked the RegistryClient to just take the TLSConfiguration.

Also moved the `makeEnvironmentAwareTLSConfiguration` as we already need it twice in `containerization`

dcantah: Thanks!! Reviewing

sebimarkgraf: @dcantah This just blocked on available review time or should I do something?

sebimarkgraf:
> I'd move the tlsConfiguration parameter one spot up above the logger, I'd like to leave the logger as the last parameter. 

Moved the `tlsConfiguration` above the logger :)

sebimarkgraf:
@dcantah Sorry for pinging again, but I am currently cleaning up my PRs before moving into the next year.

Is there anything left for me to do or is this just waiting for a good time to merge?

dcantah:
@sebimarkgraf Nothing at all, sorry! Just been a bit backed up. Will check this in after green CI!

dcantah:
@sebimarkgraf Sorry, do you mind rebasing on main and repushing 😅. That should fix the CI failure

dcantah:
Lovely, first time trying the rebase UI but we require verified signatures so that just doesn't work 😄. Just rebase and force push here and we should be good to check in

sebimarkgraf:
Rebase is done :)

PR #408, opened 2025-11-13
LinuxContainer: Keep reference to vended execs

dcantah:
This change is aimed at making forgetting to call .delete() on a LinuxProcess less destructive than it can be. Because Virt.framework invalidates any vsock fds it vended if the vm is stopped, trying to perform some operations on the grpc client through any of the process methods could trigger an ebadf, which NIO asserts on. This keeps a reference to the execs and deletes all of them for you once the container dies. I still think leaving .delete a public method is useful as otherwise the stdio fds are left open, but cleanup should occur all in one place now if you don't care about this.

This additionally:

1. Fixes two of our tests that forgot to delete() an exec.
2. Adds two new tests to verify that process.delete() is now idempotent, and we don't need to call delete().

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -127,17 +127,20 @@ public final class LinuxContainer: Container, Sendable {
             let vm: any VirtualMachineInstance
             let process: LinuxProcess
             let relayManager: UnixSocketRelayManager
+            var vendedProcesses: [String: LinuxProcess]
 
             init(_ state: CreatedState, process: LinuxProcess) {
                 self.vm = state.vm
                 self.relayManager = state.relayManager
                 self.process = process
+                self.vendedProcesses = [:]
             }
 
             init(_ state: PausedState) {
                 self.vm = state.vm
                 self.relayManager = state.relayManager
                 self.process = state.process
+                self.vendedProcesses = state.vendedProcesses
             }
         }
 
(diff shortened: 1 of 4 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -626,16 +635,23 @@ extension LinuxContainer {
...
-                vm: state.vm,
-                logger: self.logger
+                vm: startedState.vm,
+                logger: self.logger,
+                onDelete: { [weak self] in
+                    await self?.removeProcess(id: id)
+                }
             )
+
+            startedState.vendedProcesses[id] = process
```
dkovba, line 652: Should this be done before the code below?
```
            let process = LinuxProcess(
                ...
                onDelete: { [weak self] in
                    await self?.removeProcess(id: id)
                }
```

dkovba (approved):

Sources/Containerization/LinuxContainer.swift
dcantah: The callback has zero chance of firing as the process isn't started in the constructor, so to me it doesn't raise any alarms

dcantah: Wait, perhaps I'm not understanding what you're proposing though?

dcantah: Going to check this in but let me know in the AM!

dkovba: If `onDelete` can't be executed sooner, there is no problem. But generally, it's safer to do this in the reversed order.

PR #411, opened 2025-11-15
ManagedProcess: Capture vmexec stderr and convert to Containerization Error

### LINKED ISSUE #277: [Request]: vmexec: Errors should turn into real errors

Today if something explodes in vmexec I believe it simply just gets logged to stderr and the process is marked as exiting with ec 1. We should have some mechanism to turn the error text into a real error (CZError) and bubble this up.

### Code of Conduct

RahulThennarasu:
Fixes #277

When vmexec fails, it logs to stderr and exits with code 1. Previously, the error details were lost. This change captures stderr and converts it into a proper ContainerizationError.

**Changes:**
- Added stderr pipe to capture vmexec output
- Modified setExit() to convert stderr to ContainerizationError on non-zero exit
- Added optional error field to ExitStatus struct

**Testing:** All 167 tests pass.

### DIFF
--- a/Sources/Integration/ContainerTests.swift
+++ b/Sources/Integration/ContainerTests.swift
@@ -1227,4 +1227,23 @@ extension IntegrationSuite {
             throw error
         }
     }
+
+    func testNonExistentBinary() async throws {
+        let id = "test-non-existent-binary"
+
+        let bs = try await bootstrap(id)
+        let container = try LinuxContainer(id, rootfs: bs.rootfs, vmm: bs.vmm) { config in
+            config.process.arguments = ["foo-bar-baz"]
+            config.bootLog = bs.bootLog
+        }
+
+        try await container.create()
+        do {
+            try await container.start()
+        } catch {
+            return
+        }
+        try await container.stop()
+        throw IntegrationError.assert(msg: "container start should have failed")
+    }
 }
--- a/vminitd/Sources/vminitd/ManagedProcess.swift
+++ b/vminitd/Sources/vminitd/ManagedProcess.swift
@@ -63,6 +63,7 @@ final class ManagedProcess: Sendable {
     private let owningPid: Int32?
     private let ackPipe: Pipe
     private let syncPipe: Pipe
+    private let errorPipe: Pipe
     private let terminal: Bool
     private let bundle: ContainerizationOCI.Bundle
     private let cgroupManager: Cgroup2Manager?
@@ -95,6 +96,10 @@ final class ManagedProcess: Sendable {
         try ackPipe.setCloexec()
         self.ackPipe = ackPipe
 
+        let errorPipe = Pipe()
+        try errorPipe.setCloexec()
+        self.errorPipe = errorPipe
+
         let args: [String]
         if let owningPid {
             args = [
@@ -114,6 +119,7 @@ final class ManagedProcess: Sendable {
             extraFiles: [
                 syncPipe.fileHandleForWriting,
                 ackPipe.fileHandleForReading,
+                errorPipe.fileHandleForWriting,
             ]
         )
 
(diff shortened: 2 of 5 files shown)


dcantah:
This unfortunately is not what we'd want here. The command already has its stderr redirected to a set of pipes we'll read from and then relay to some vsock sockets to the host. Ideally the solution here is just an extra pipe or file that vmexec writes any errors to that we check if any of the setup stages fail.

RahulThennarasu:
Thanks for letting me know! I've switched to using a dedicated error pipe instead of stderr. 

vmexec writes errors to FD 5 before exiting and ManagedProcess reads from that pipe and converts to ContainerizationError.
It doesn't touch the existing stderr relay. Let me know if this approach works better.

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -158,6 +168,16 @@ extension ManagedProcess {
...
             try command.start()
+
+            if let errorPipe = $0.errorPipe {
+                let errorReadHandle = errorPipe.fileHandleForReading
+                Task { [weak self] in
+                    if let data = try? errorReadHandle.readToEnd() {
+                        self?.state.withLock { $0.errorData = data }
+                    }
+                } 
+            }
```
dcantah: What I had in mind was we'd have a big do catch block around all of this setup, and if any of the syncPipe, ackPipe or any other setup failed, we'd try and read any error data from the error pipe. vmexec with your new setup should write any errors it encountered always, so if something went awry here we'd read it in the catch block. We want ManagedProcess.start() to throw an error if something went wrong, not store some error text in the background.

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -39,6 +39,7 @@ final class ManagedProcess: Sendable {
     struct ExitStatus {
         var exitStatus: Int32
         var exitedAt: Date
+        var error: ContainerizationError?
```
dcantah: See here first https://github.com/apple/containerization/pull/411#discussion_r2530330692, but we don't need to store this. What the issue was trying to convey is we should have `ManagedProcess.start` return an error if any of the setup didn't actually work. It doesn't seem like we even do anything with this data here in the rest of the change. The error field on ExitStatus isn't propagated to the host.

dcantah:
@RahulThennarasu also, please message on a given issue if you want to pick up the work so we can assign you. Usually if someone is already assigned they may have some context that is useful still if they haven't gotten around to an actual change.

RahulThennarasu:
Now the start() wraps everything in do-catch, reads the error pipe in the catch block, and throws immediately with vmexec's error message. I removed the ExitStatus.error field and background task as well.

I also will comment on issues before picking them up going forward.

dcantah:
To test this, can you add a new integration test to ContainerTests.swift that tries to spawn a non-existent binary in the image (foo-bar-baz or any other binary name we know isn't valid) and confirm that LinuxContainer.start returns an error?

RahulThennarasu:
I fixed the unused result warning and added the test. The test tries to spawn foo-bar-baz and verifies that start() throws a ContainerizationError with the vmexec error message in it.

Sources/Integration/ContainerTests.swift
```diff
@@ -1227,4 +1227,32 @@ extension IntegrationSuite {
...
+            // Verify the error message contains information from vmexec
+            guard error.message.contains("vmexec error") || error.message.contains("failed to find target executable") else {
+                throw IntegrationError.assert(msg: "error message should contain vmexec error details, got: \(error.message)")
+            }
+            // Success if we got the expected error with vmexec details
+        } catch {
+            // Re-throw if it's not a ContainerizationError
+            throw error
+        }
+    }
```
dcantah, line 1248: I'd prefer we don't try and match on error text, as it's not really a contract we have at all. Just seeing that start throws an error is fine. So:

```swift
try await container.create()
do {
    try await container.start()
} catch {
    return
}
try await container.stop()
throw IntegrationError.assert(msg: "container start should have failed")
```

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -144,130 +149,153 @@ final class ManagedProcess: Sendable {
         self.terminal = stdio.terminal
         self.bundle = bundle
         self.state = Mutex(State(io: io))
+        self.state.withLock { state in
+            state.errorPipe = errorPipe
+        }
```
dcantah: Just add a new errorPipe field to the constructor so we can do `Mutex(State(io: io, errorPipe: errorPipe))` instead of this.

RahulThennarasu:
hi, I simplified the test and I also moved errorPipe into the State constructor.

Sources/Integration/ContainerTests.swift
jglogan: In a test, matching against the text can be necessary to ensure branch coverage. If we change the text, the test breaks as we should and then we figure out what best to do next.

Actual client code shouldn't make decisions based on message.

EDIT: Never mind for this case – we're not really checking branch coverage with the above code.

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -143,131 +149,151 @@ final class ManagedProcess: Sendable {
         self.command = command
         self.terminal = stdio.terminal
         self.bundle = bundle
-        self.state = Mutex(State(io: io))
+        self.state = Mutex(State(io: io, errorPipe: errorPipe))
```
dcantah: Sorry, I kind of inadvertently recommended this, but I'd forgotten how we'd done the other two pipes. Can you just instantiate the error pipe in the constructor, and then just have it be a member variable of this type just like syncPipe and ackPipe. I don't think we need to lock around it.

vminitd/Sources/vminitd/ManagedProcess.swift
```diff
@@ -143,131 +149,151 @@ final class ManagedProcess: Sendable {
...
-                throw ContainerizationError(.internalError, message: "no PID data from sync pipe")
-            }
+                // Start the underlying process.
+                try command.start()
 
-            guard piddata.count == size else {
-                throw ContainerizationError(.internalError, message: "invalid payload")
-            }
+                defer {
+                    try? self.ackPipe.fileHandleForWriting.close()
```
dcantah, line 170: Related to my other latest comment, if we just have errorPipe as a variable on ManagedProcess (not in the state mutex) add a close for it here.

dcantah:
Two little bits of feedback, and then I think this looks good

RahulThennarasu:
yep, I moved errorPipe out of State and made it a member variable like syncPipe and ackPipe.

PR #415, opened 2025-11-18
Prevent recursion attacks in EXT4Formatter.unlink

wlan0:
This issue doesn't affect any of our existing products. This is a preemptive fix for downstream consumers of EXT4.format where, in some platforms, leading `//` in the path could get resolved into a FileTree that looks like this

```sh
/ 
 └──  / 
 └── usr
```

### DIFF
--- a/Sources/ContainerizationEXT4/EXT4+Formatter.swift
+++ b/Sources/ContainerizationEXT4/EXT4+Formatter.swift
@@ -199,42 +199,63 @@ extension EXT4 {
         //
         // - `path`: The `FilePath` specifying the path of the file or directory to delete.
         public func unlink(path: FilePath, directoryWhiteout: Bool = false) throws {
-            guard let pathPtr = self.tree.lookup(path: path) else {
+            guard let pathPtr = self.tree.lookup(path: path),
+                let parentPtr = self.tree.lookup(path: path.dir)
+            else {
                 // We are being asked to unlink something that does not exist. Ignore
                 return
             }
             let pathNode = pathPtr.pointee
             let inodeNumber = Int(pathNode.inode) - 1
             let pathInodePtr = self.inodes[inodeNumber]
-            var pathInode = pathInodePtr.pointee
+            let pathInode = pathInodePtr.pointee
 
             if directoryWhiteout && !pathInode.mode.isDir() {
                 throw Error.notDirectory(path)
             }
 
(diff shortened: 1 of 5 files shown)


Sources/ContainerizationOS/Socket/Socket.swift
```diff
@@ -329,7 +329,7 @@ extension Socket {
 
         var cmsgBuf = [UInt8](repeating: 0, count: Int(CZ_CMSG_SPACE(Int(MemoryLayout<Int32>.size))))
         msg.msg_control = withUnsafeMutablePointer(to: &cmsgBuf[0]) { UnsafeMutableRawPointer($0) }
-        msg.msg_controllen = socklen_t(cmsgBuf.count)
+        msg.msg_controllen = numericCast(cmsgBuf.count)
```
crosbymichael, line 333: What is this change for? Looks to be causing a merge conflict as well.

wlan0: `numericCast` takes care of converting cmsgBuf.count (Int) to whatever type is expected by the platform

msg.msg_controllen is 

- `socklen_t` on macOS
- `size_t` on Linux

Sources/ContainerizationEXT4/EXT4+Formatter.swift
```diff
@@ -201,21 +201,42 @@ extension EXT4 {
...
+            let pathInode = pathInodePtr.pointee
 
             if directoryWhiteout && !pathInode.mode.isDir() {
                 throw Error.notDirectory(path)
             }
 
-            for childPtr in pathNode.children {
-                try self.unlink(path: path.join(childPtr.pointee.name))
+            // Iterative breath-first traversal of the FileTree to prevent recursion attacks
+            var queue: [(parent: Ptr<FileTree.FileTreeNode>?, entry: Ptr<FileTree.FileTreeNode>)] = pathNode.children.map { (pathPtr, $0) }
```
jglogan: `parent` need not be optional; there's no way it can be `nil` right now.

jglogan (changes requested):

dkovba (approved):

crosbymichael:
@wlan0 is this still needed? If so we need a rebase.

jglogan (approved):

PR #418, opened 2025-11-19
Vminitd: Add pause command

dcantah:
Due to us supporting a pod type now, and pid ns sharing being quite a common thing for pods, lets add a pause container like command to vminitd to eventually enable pid ns sharing between containers in our variant of a pod.

This changes vminitd slightly to have pause and init (default) commands as it seemed simpler than creating a whole new binary to include in the guest image.

### DIFF
--- a/vminitd/Sources/vminitd/PauseCommand.swift
+++ b/vminitd/Sources/vminitd/PauseCommand.swift
@@ -1,0 +1,59 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Containerization project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


vminitd/Sources/vminitd/PauseCommand.swift
```diff
@@ -0,0 +1,61 @@
...
+        sigintSource.resume()
+
+        let sigtermSource = DispatchSource.makeSignalSource(signal: SIGTERM)
+        sigtermSource.setEventHandler {
+            log.info("Shutting down, got SIGTERM")
+            Musl.exit(0)
+        }
+        sigtermSource.resume()
+
+        let sigchldSource = DispatchSource.makeSignalSource(signal: SIGINT)
```
dcantah: whoops hang on, screwed this up in latest push

crosbymichael:
This should pass with a rebase now.

PR #419, opened 2025-11-19
Add images prune command to reclaim disk space

### LINKED ISSUE #417: [Bug]: `ImageStore.prune()` doesn't remove unreferenced images
### I have done the following

### Steps to reproduce

`ImageStore.prune()` (which calls `ImageStore._prune()`) collects blob digests from all images in the store and only deletes blobs not referenced by any image. This means it performs content-store garbage collection but never removes complete images, even if those images aren't used by any containers. We need a way to pass in active image references (from containers) so the method can remove unreferenced images and their blobs. 

The `container image prune` command calls this method and advertises that it removes "unreferenced and dangling images", but because `ImageStore.prune()` only removes orphaned blobs, users see 0 B reclaimed even when they have many unused images taking up disk space.

RahulThennarasu:
Fixes #417 

Added a `prune` subcommand to `cctl images` that calls `ImageStore.prune()` to remove unreferenced blobs from the content store. Displays number of blobs deleted and total space reclaimed

**Testing:**
Verified the command works by pulling alpine:latest, deleting it, then running prune which reclaimed 29.8 MB from 56 unreferenced blobs.

**Usage:**
```bash
cctl images prune
```

### DIFF
--- a/Sources/cctl/ImageCommand.swift
+++ b/Sources/cctl/ImageCommand.swift
@@ -35,6 +35,7 @@ extension Application {
                 Push.self,
                 Save.self,
                 Load.self,
+                Prune.self,
             ]
         )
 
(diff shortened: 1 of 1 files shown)


dcantah:
Howdy! Unfortunately the issue wasn't regarding wanting a cctl command for image prune, it was about our prune logic not really being rock solid. I don't think we want a prune command on cctl, but we really appreciate the contribution!

PR #420, opened 2025-11-20
Fix ImageStore.prune() behavior to actually remove images

### LINKED ISSUE #417: [Bug]: `ImageStore.prune()` doesn't remove unreferenced images

realrajaryan:
- Fixes #417.

Rename `_prune()` to `cleanupOrphanedBlobs()` to clarify what it actually does, and remove `prune()` method as we'll do all that logic in container directly.

### DIFF
--- a/Sources/Containerization/Image/ImageStore/ImageStore.swift
+++ b/Sources/Containerization/Image/ImageStore/ImageStore.swift
@@ -129,43 +129,6 @@ extension ImageStore {
         }
     }
 
-    /// Perform a garbage collection in the underlying `ContentStore` that is managed by the `ImageStore`.
-    ///
-    /// - Parameters:
-    ///   - keepingReferences: List of image references to keep. All images NOT in this list will be removed.
-    ///
-    /// - Returns: Returns a tuple of `(deleted, freed)`.
-    ///   `deleted` :  A  list of the names of the content items that were deleted from the `ContentStore`,
-    ///   `freed` : The total size of the items that were deleted.
-    @discardableResult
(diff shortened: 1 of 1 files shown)


realrajaryan:
making cleanupOrphanedBlobs public because we'll need it in container as well

jglogan:
Could this be handled without any changes to Containerization? That is to say:

* `image prune` cleans only orphaned blobs (those with no image references)
* `image prune --all` the API server first finds all images having no container references , runs `rm` on all of those, then cleans orphaned blobs

realrajaryan:
> See above comment. I guess we would need to make cleanOrphanedBlobs() public for `image prune` but could we do the image pruning completely in container-apiserver?

Yeah like we discussed, we basically rename `_prune` to `cleanupOrphanedBlobs()` to clarify what it actually does, and I'll handle the rest of the pruning logic in container.

dcantah (approved):

jglogan (approved):

PR #431, opened 2025-11-24
Implement pdeathSignal handling in Runc

gautamrajur:
Add support for parent death signal (pdeathSignal) to ensure child processes receive a signal when the parent process dies. This addresses the FIXME comment in Runc.execute().

## Changes
- Add pdeathSignal field to exec_command_attrs C struct
- Implement prctl(PR_SET_PDEATHSIG) in child process handler (Linux only)
- Expose pdeathSignal through Command.Attrs Swift API
- Wire up pdeathSignal in Runc.execute() to remove FIXME

## Implementation Details
The implementation uses Linux-specific prctl() to set the parent death signal, ensuring proper cleanup when parent processes terminate. The feature is conditionally compiled for Linux only, maintaining compatibility with other platforms.

## Testing
This change maintains backward compatibility as pdeathSignal defaults to nil/0, meaning existing code will continue to work without modification.

### DIFF
--- a/Sources/CShim/exec_command.c
+++ b/Sources/CShim/exec_command.c
@@ -32,6 +32,9 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#if defined(__linux__)
+#include <sys/prctl.h>
+#endif
 
 #include "exec_command.h"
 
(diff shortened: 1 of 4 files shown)


dcantah:
@gautamrajur Hey can you sign your commit (gpg, ssh) and repush? This looks good to me after.

gautamrajur:
I hope this looks good now

Issue #442, opened 2025-12-09
[Request]: Capabilities support


dcantah:
### Feature or enhancement request details

We don't support capabilities today, so you just get everythang. We should support a similar interface to the OCI spec, and we'll need a library that sets things correctly. I'm sure we'll get it wrong originally and will need some iteration :)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #444, opened 2025-12-10
Add capabilities support

dcantah:
Closes https://github.com/apple/containerization/issues/442

This adds capabilities support to LinuxContainer via a new surface in ContainerizationOS + some C wrappers.

### DIFF
--- a/Sources/Containerization/LinuxProcessConfiguration.swift
+++ b/Sources/Containerization/LinuxProcessConfiguration.swift
@@ -17,6 +17,116 @@
 import ContainerizationOCI
 import ContainerizationOS
 
+/// User-friendly Linux capabilities configuration
+public struct LinuxCapabilities: Sendable {
+    /// Capabilities that define the maximum set of capabilities a process can have
+    public var bounding: [CapabilityName] = []
+    /// Capabilities that are actually in effect for the current process
+    public var effective: [CapabilityName] = []
+    /// Capabilities that can be inherited by child processes
+    public var inheritable: [CapabilityName] = []
+    /// Capabilities that are currently permitted for the process
+    public var permitted: [CapabilityName] = []
+    /// Capabilities that are preserved across execve() calls
+    public var ambient: [CapabilityName] = []
+
+    /// Grant all capabilities
+    public static let allCapabilities = LinuxCapabilities(
+        bounding: CapabilityName.allCases,
+        effective: CapabilityName.allCases,
+        inheritable: CapabilityName.allCases,
+        permitted: CapabilityName.allCases,
+        ambient: CapabilityName.allCases
+    )
+
+    /// Default configuration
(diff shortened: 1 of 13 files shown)


Sources/ContainerizationOCI/Spec.swift
```diff
@@ -194,25 +195,51 @@ public struct Process: Codable, Sendable {
...
+        ambient: [String]? = nil
     ) {
         self.bounding = bounding
         self.effective = effective
         self.inheritable = inheritable
         self.permitted = permitted
         self.ambient = ambient
     }
+
+    public init(from decoder: Decoder) throws {
```
crosbymichael, line 226: Why did you have to implement these codable funcs? Did the defaults not work anymore?

Sources/Containerization/LinuxProcessConfiguration.swift
```diff
@@ -31,6 +141,8 @@ public struct LinuxProcessConfiguration: Sendable {
     public var user: ContainerizationOCI.User = .init()
     /// The rlimits for the container process.
     public var rlimits: [POSIXRlimit] = []
+    /// The Linux capabilities for the container process.
+    public var capabilities: LinuxCapabilities = .defaultOCICapabilities
```
crosbymichael: Is setting the default caps a breaking change for users now?

dcantah: It is, I teetered between giving all caps, or the OCI defaults here. Thoughts?

Sources/ContainerizationOCI/Spec.swift
dcantah: The oci spec lists them as optional which from what I've found in Go land means they actually expect a nil slice if its not provided (got bit by this from runc expecting some slice fields on Mount to be nil instead of empty slices)

dcantah: Any omitEmpty slices should be given the same treatment here, but we're kind of going one type at a time 😄

Sources/Containerization/LinuxProcessConfiguration.swift
crosbymichael: For this initial change I would probably keep the current functionality.  Give people time, including `container` to set these explicitly, and then we can find a path for a better default post merge.

Sources/ContainerizationOCI/Spec.swift
crosbymichael: ahhh see

Sources/Containerization/LinuxProcessConfiguration.swift
dcantah: Whenever container would ingest this change I was going to just set it to .allCapabilities. But we can just make the default that here and avoid having to do anything for now

dcantah: Swapped to all by default

PR #447, opened 2025-12-12
Feat: support configuring the disk storage size

karenheckel:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
- For support of specifying a default disk capacity.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerizationOS/Command.swift
+++ b/Sources/ContainerizationOS/Command.swift
@@ -183,7 +183,9 @@ extension Command {
 
         let set = try createFileset()
         defer {
-            try? set.null.close()
+            for nullHandle in set.nullHandles {
+                try? nullHandle.close()
+            }
         }
         var fds = [Int32](repeating: 0, count: set.handles.count)
         for (i, handle) in set.handles.enumerated() {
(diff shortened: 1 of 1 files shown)


dcantah:
Hi @karenheckel, thanks for the contribution! Just truncating the file won't have the filesystem inside updated to believe it has {insert new size} available to it unfortunately so this won't work as is. The approach I'd been thinking of for this problem, although this would be a tiny bit complex, is to leave the actual block containing the image contents alone (through the library you can choose whatever size you want for this already) and instead provide a secondary block that we'd create on the fly of the requested size. We'd attach both the container image block (as we already do today) and this new block of the requested size to the VM, mount both of them, and then union the two with [overlayfs](https://docs.kernel.org/filesystems/overlayfs.html) in the guest, where the upperdir would be the empty block of the requested size. The rationale for this is that in the container you'd now see you have {requested size} available to write. This works both ways, if you want to give a smaller filesystem size this should work completely fine, as well as a much larger size than our images are (512GiB) today as well. Then when the container exits we can just remove the block we made for this little dance.

jglogan:
@karenheckel I need to unburden myself from a couple other tasks but early next week let me look at your container PR and your latest comments there, and we can work out how to move the enhancement forward.

dcantah:
Hi @karenheckel this was solved via https://github.com/apple/containerization/pull/533. It hasn't been integrated into https://github.com/apple/container however if you'd like to tackle that side

karenheckel:
Hi @dcantah! I can tackle it on the container side. I should be able to start taking a look next week. Do you have any pointers for it given what's already been done?

PR #463, opened 2025-12-31
Agent: Add single file copying

dcantah:
Add a streaming rpc to copy single files in and out. This can easily be extended to support copying directories, but we need the guest agent to be able to tar/untar, and we'll need to figure out what that model is.

### DIFF
--- a/Sources/Containerization/VirtualMachineAgent.swift
+++ b/Sources/Containerization/VirtualMachineAgent.swift
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2025 Apple Inc. and the Containerization project authors.
+// Copyright © 2025-2026 Apple Inc. and the Containerization project authors.
 //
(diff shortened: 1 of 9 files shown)


Sources/Containerization/VirtualMachineAgent.swift
```diff
@@ -46,6 +46,26 @@ public protocol VirtualMachineAgent: Sendable {
...
+        from source: URL,
+        to destination: URL,
+        mode: UInt32,
+        createParents: Bool,
+        chunkSize: Int,
+        progress: ProgressHandler?
+    ) async throws
+
+    /// Copy a file from the guest to the host.
+    func copyOut(
```
jglogan, line 62: Copy out always creates parent directories on the host whereas copy in allows choosing what happens on the guest?

dcantah: Let me make it symmetrical

jglogan (approved):

PR #468, opened 2026-01-03
Adds MACAddress type.

jglogan:
- Simplifies processing and validation of MAC addresses for apple/container#1005, where we need to create MAC addresses and corresponding link local IPv6 addresses.
- Adds Codable for IPAddress and CIDR for apple/container#1006, so we can parse and relay IP addresses regardless of address family for port forwards.

### DIFF
--- a/Sources/ContainerizationExtras/AddressError.swift
+++ b/Sources/ContainerizationExtras/AddressError.swift
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2025 Apple Inc. and the Containerization project authors.
+// Copyright © 2026 Apple Inc. and the Containerization project authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/Sources/ContainerizationExtras/IPv4Address.swift
+++ b/Sources/ContainerizationExtras/IPv4Address.swift
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2025 Apple Inc. and the Containerization project authors.
+// Copyright © 2025-2026 Apple Inc. and the Containerization project authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/Sources/ContainerizationExtras/IPv6Address+Parse.swift
+++ b/Sources/ContainerizationExtras/IPv6Address+Parse.swift
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2025 Apple Inc. and the Containerization project authors.
+// Copyright © 2025-2026 Apple Inc. and the Containerization project authors.
 //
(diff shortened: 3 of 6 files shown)


Sources/ContainerizationExtras/MACAddress.swift
```diff
@@ -0,0 +1,256 @@
...
+    ///   The most significant 16 bits of the value are ignored.
+    @inlinable
+    public init(_ value: UInt64) {
+        self.value = value & 0x0000_ffff_ffff_ffff
+    }
+
+    /// Creates an MACAddress from a string representation.
+    ///
+    /// - Parameter string: The IPv4 address string in dotted decimal notation (e.g., "192.168.1.1")
+    /// - Throws: `IPAddressError.unableToParse` if the string is not a valid IPv4 address
```
dcantah: Doesn't seem to be related to this type

Sources/ContainerizationExtras/MACAddress.swift
```diff
@@ -0,0 +1,256 @@
...
+    ///   or dashes
+    /// - No whitespace characters
+    /// - Only hexadecimal digits and colons allowed
+    ///
+    /// ## Examples
+    /// ```swift
+    /// MACAddress.parse("01:23:45:67:89:ab") // Returns: 0x00000123456789ab
+    /// MACAddress.parse("01-23-45-67-89-AB") // Returns: 0x00000123456789ab
+    /// MACAddress.parse("00:00:00:00:00:00") // Returns: 0x0000000000000000
+    /// MACAddress.parse("ff:ff:ff:ff:ff:ff") // Returns: 0xffffffffffffffff
```
dcantah: This seems like it's not ignoring the top 16 bits

Sources/ContainerizationExtras/MACAddress.swift
```diff
@@ -0,0 +1,256 @@
...
+        guard !s.isEmpty, s.count == 17 else {
+            throw IPAddressError.unableToParse
+        }
+
+        // MAC addresses should only contain ASCII hex digits and dots
+        let utf8 = s.utf8
+        for byte in utf8 {
+            // ASCII whitespace: space(32), tab(9), newline(10), return(13)
+            if byte == 32 || byte == 9 || byte == 10 || byte == 13 {
+                throw IPAddressError.unableToParse
```
dcantah: Oh maybe my prior comment was only partially wrong, but I would expect we return some MacAddressError instead

jglogan: I changed IPAddressError to AddressError instead of creating a redundant error type, but I can create a separate MACAddressError for this type if we think it's useful.

Sources/ContainerizationExtras/MACAddress.swift
jglogan: Fixed, thanks!

Sources/ContainerizationExtras/MACAddress.swift
jglogan: Fixed

PR #471, opened 2026-01-09
Allow filtering container statistics

dcantah:
It's possible a user doesn't want the full stats list, and only wants cpu/mem etc. This plumbs through the ability to filter to only what is requested. This, while we're already here, adds in memory.event output to the stats list. For that specifically, I think eventually we may want a streaming variant of this so you can get alerted of changes in the file immediately instead of polling/one off reads, but this is useful for now.


dcantah:
Ok, I changed my mind. I think the events should just come along the ride in the statistics call we already have. However, I think we should allow filtering what stats a user actually wants in the call too. I'll change this pr

Sources/Containerization/ContainerStatistics.swift
```diff
@@ -174,4 +177,51 @@ public struct ContainerStatistics: Sendable {
             self.transmittedErrors = transmittedErrors
         }
     }
+
+    /// Memory event counters from cgroup2's memory.events file.
+    public struct MemoryEventStatistics: Sendable {
```
egernst, line 182: Proto defines oom_group_kill - should we add here, or remove from proto?

dcantah: I'd prefer to leave it for now, it's mostly harmless and we're the only consumers of this info. We don't have a way today to set the oom group kill file so the value should always be 0

egernst (approved):
needs a rebase. lgtm.

PR #473, opened 2026-01-10
Fix: resolve relative executable paths against workingDirectory in vmexec

ParkSeongGeun:
## Summary

Fixes the root cause for apple/container#962.

This PR fixes an issue where relative paths provided as an entrypoint or command (e.g., `./uname`) failed to execute. The root cause was that executable paths were being validated before the working directory was changed (`chdir`) in the guest VM executor, leading to a "failed to find target executable" error.

## Motivation and Context
In the current `vmexec` implementation, the file existence check occurs before applying the OCI `workingDirectory`. Following the "cleaner approach" suggested by maintainers in apple/container#987, this change ensures that the executor itself handles relative path resolution internally.

**Key Changes:**
- **Relative Path Detection**: The executor now detects paths containing `/` or starting with `.`
- **CWD-based Resolution**: It resolves these relative paths against the OCI-specified `workingDirectory` before performing the binary existence check.
- **Path Normalization**: Utilizes `URL.standardized.path` to ensure a clean absolute path is passed to the execution system.
- **Bare Command Preservation**: Explicitly preserves bare command resolution (e.g., `ls`, `uname`) via the system `PATH` by checking for the absence of slashes.

This aligns the executor's behavior with industry standards seen in Docker and Podman.

### Integration Tests
- **Primary Scenario**: Executed `container run --workdir /bin --entrypoint ./uname alpine`.
- **Result**: Successfully resolved to `/bin/uname` inside the VM and returned `Linux`.
- **Regression**: Confirmed that absolute paths (`/bin/ls`) and bare commands (`ls`) continue to function correctly via system `PATH`.

### Cross-Repository Coordination
- **CLI Compatibility**: Verified that this fix works seamlessly with the `container` CLI.
- **Unit Test Coordination**: This change is coordinated with the newly added **`testProcessEntrypointRelativePathPassthrough`** in the `apple/container` repository (ParserTest.swift). 
- **Proof of Flow**: Confirmed that when the CLI passes a relative path untouched, this updated `vmexec` correctly handles the resolution internally, keeping the OCI bundle configuration clean and logic centralized in the executor.

### DIFF
--- a/vminitd/Sources/vmexec/vmexec.swift
+++ b/vminitd/Sources/vmexec/vmexec.swift
@@ -68,20 +68,33 @@ extension App {
             throw App.Errno(stage: "exec", info: "process args cannot be empty")
         }
 
-        // lookup executable
-        let path = Path.findPath(currentEnv) ?? Path.getCurrentPath()
-        guard let resolvedExecutable = Path.lookPath(process.args[0], path: path) else {
-            throw App.Failure(message: "failed to find target executable \(process.args[0])")
+        let executableArg = process.args[0]
+        let resolvedExecutable: URL
+
+        if executableArg.contains("/") {
+            if executableArg.hasPrefix("/") {
+                resolvedExecutable = URL(fileURLWithPath: executableArg)
+            } else {
+                resolvedExecutable = URL(fileURLWithPath: process.cwd).appendingPathComponent(executableArg).standardized
+            }
+
+            guard FileManager.default.fileExists(atPath: resolvedExecutable.path) else {
+                throw App.Failure(message: "failed to find target executable \(executableArg)")
+            }
+        } else {
+            let path = Path.findPath(currentEnv) ?? Path.getCurrentPath()
+            guard let found = Path.lookPath(executableArg, path: path) else {
+                throw App.Failure(message: "failed to find target executable \(executableArg)")
+            }
+            resolvedExecutable = found
         }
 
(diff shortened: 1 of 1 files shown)


vminitd/Sources/vmexec/vmexec.swift
```diff
@@ -68,20 +68,33 @@ extension App {
...
+        } else {
+            let path = Path.findPath(currentEnv) ?? Path.getCurrentPath()
+            guard let found = Path.lookPath(executableArg, path: path) else {
+                throw App.Failure(message: "failed to find target executable \(executableArg)")
+            }
+            resolvedExecutable = found
         }
 
-        let executable = strdup(resolvedExecutable.path())
+        guard FileManager.default.fileExists(atPath: resolvedExecutable.path) else {
```
jglogan: If we go through the path at L81, then the file should already exist, right? Could this move into the end of the block at L75-79?

jglogan (commented):
Looks good, see the comment, just a nit.

vminitd/Sources/vmexec/vmexec.swift
ParkSeongGeun: Yes. This was the inefficient part of checking twice. I move this part in the end of the block L75-79. @jglogan

jglogan:
@ParkSeongGeun Thanks! I see you've already updated the container PR so that we're still testing it there, was going to ask you to do that 😄

PR #474, opened 2026-01-13
_ContainerizationTar: Add new native tar reader/writer

dcantah:
Swift does not have a native tar reader/writer, and we'd really like to avoid linking more (libarchive) libraries to vminitd if we can get away with it. For copying directories into/out of containers tar is fairly nice as it's a simple way to preserve everything you need to be able to reassemble the directory on the receiving end. Due to this, I decided to write a somewhat simple tar reader/writer solely for this purpose. Unfortunately, there's quite a lot of work to get vminitd/ to compile on macOS without the static linux SDK, so it was making unit testing these additions quite a pain, so for now the new work lives in _ContainerizationTar (where the underscore is trying to denote that this is rather experimental..).

This change aims to add a simple tar reader and writer with support for pax extended headers (for long file names and > 8 GiB files). Because its intended purpose is in a scenario where we own both the creator and ingestor, the reader does NOT handle every case, but it is good at unarchiving the archives the library has made :)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "c82be4e21117351bb3f942869ce90d35dcd0dd0223dc1c49ce7a56b52709e836",
+  "originHash" : "e5fa0e8b0e9dab4b79c924cd2c585e41bb516d58ef0af8f1b3a1d1a4a7d9810d",
   "pins" : [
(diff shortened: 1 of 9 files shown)


dcantah:
libarchive made it into the static linux sdk, so closing this!

PR #487, opened 2026-01-22
LinuxContainer/LinuxPod: Single file support

### LINKED ISSUE #79: Single file mount support
It would be nice to support sharing in single files instead of being bound by it being a directory.

Virtiofs doesn't support sharing in single files, so we'd need to get a bit smart. One option is to share in the parent directory of the file, and then bind mount the single file into the containers mount ns. The mount setup code would need to start doing a stat on every mount to ensure it's a file and not a directory and take appropriate action, but that's not too worrying.

dcantah:
Fixes #79

This adds support for single file (virtiofs based) mounts to the two main container types. This is transparent, and doesn't need to be setup by a user, so there is no visible API change. Virtiofs does not support single file mounts today, and it would be less than ideal from a security standpoint to just expose the parent directory of the file to the vm and then bind mount in the file, so we chose the following:

1. Create a temp directory the container types will manage.
2. Hardlink in the file we want.
3. Share that tempdir into the vm.
4. Finally bind mount in the file into the container.

The main goal I wanted for this is to leave the logic out of the `VirtualMachineInstance`. I really want this to just be a little dance the container types do, and leave the vm out of it. Because of that, most of the logic is shoved in a new `FileMountContext` type that does the dirty tricks of rewriting what the user asked for to be a bind mount from a temporary holding spot we'll use in the guest for these temp directories.

One of the downsides to this is today the tempdirs are solely on the main volume, so cross volume will need to copy the file to the tempdir, so writes won't get written back.

### DIFF
--- a/Sources/Containerization/FileMount.swift
+++ b/Sources/Containerization/FileMount.swift
@@ -1,0 +1,241 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the Containerization project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+#if os(macOS)
+
+import ContainerizationError
+import ContainerizationOCI
+import Foundation
+
+/// Manages single-file mounts by transforming them into virtiofs directory shares
+/// plus bind mounts.
+///
+/// Since virtiofs only supports sharing directories, mounting a single file without
+/// exposing the other potential files in that directory needs a little bit of a "hack".
+/// The one we've landed on is:
+///
+/// 1. Creating a temporary directory containing a hardlink to the file
+/// 2. Sharing that directory via virtiofs to a holding location in the guest
+/// 3. Bind mounting the specific file from the holding location to the final destination
+///
+/// This type handles all three steps transparently.
+struct FileMountContext: Sendable {
+    /// Metadata for a single prepared file mount.
+    struct PreparedMount: Sendable {
+        /// Original file path on host
+        let hostFilePath: String
+        /// Where the user wants the file in the container
+        let containerDestination: String
+        /// Just the filename
+        let filename: String
+        /// Temp directory containing the hardlinked file
+        let tempDirectory: URL
+        /// The virtiofs tag (hash of temp dir path). Used to find the AttachedFilesystem
+        let tag: String
+        /// Mount options from the original mount
+        let options: [String]
+        /// Where we mounted the share in the guest (set after mountHoldingDirectories)
+        var guestHoldingPath: String?
+    }
+
+    /// Prepared file mounts for this context
+    var preparedMounts: [PreparedMount]
+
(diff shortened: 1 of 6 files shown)


Sources/Containerization/FileMount.swift
```diff
@@ -0,0 +1,241 @@
...
+import Foundation
+
+/// Manages single-file mounts by transforming them into virtiofs directory shares
+/// plus bind mounts.
+///
+/// Since virtiofs only supports sharing directories, mounting a single file without
+/// exposing the other potential files in that directory needs a little bit of a "hack".
+/// The one we've landed on is:
+///
+/// 1. Creating a temporary directory containing a hardlink to the file
```
jglogan, line 30: This will work for source files that are regular or symlink to absolute paths, will it work if the file I'm mounting is something like `schema.sql -> ../schemas/my-schema.sql`?

OK, it looks like the hard link follows the symlink to the final inode in my test:

Say PWD is `/Users/jimmy/projects/big-database`.

There is a file `/Users/jimmy/projects/schemas/my-schema.sql` and a symlink `/Users/jimmy/projects/big-database/schema.sql -> ../schemas/my-schema.sql`.

The volume mount is `-v ${PWD}/schema.sql:/mnt/db/schema/schema.sql`.

`prepareFileMount()` creates `/var/folders/6k/tnyh0vfd07z0f9mr5cg7zs5r0000gn/T/containerization-file-mounts/e7021660-61cf-48ab-b021-9dfc878c28d2` and now we do the equivalent of:

```bash
cd /var/folders/6k/tnyh0vfd07z0f9mr5cg7zs5r0000gn/T/mounts/e7021660-61cf-48ab-b021-9dfc878c28d2
ln /Users/jimmy/projects/big-database/schema.sql schema.sql
```

Sources/Containerization/FileMount.swift
```diff
@@ -0,0 +1,241 @@
...
+/// This type handles all three steps transparently.
+struct FileMountContext: Sendable {
+    /// Metadata for a single prepared file mount.
+    struct PreparedMount: Sendable {
+        /// Original file path on host
+        let hostFilePath: String
+        /// Where the user wants the file in the container
+        let containerDestination: String
+        /// Just the filename
+        let filename: String
```
jglogan, line 43: Just the filename on the host or the guest? Is there a case where it is different from both? If not, why store it when it creates the possibility of an invariant being broken by an invalid init()?

Sources/Containerization/FileMount.swift
```diff
@@ -0,0 +1,241 @@
...
+            withIntermediateDirectories: true
+        )
+
+        // Hardlink the file (falls back to copy if cross-filesystem)
+        let destURL = tempDir.appendingPathComponent(filename)
+        do {
+            try FileManager.default.linkItem(at: resolvedSource, to: destURL)
+        } catch {
+            // Hardlink failed. Fall back to copy
+            try FileManager.default.copyItem(at: resolvedSource, to: destURL)
```
jglogan, line 154: We might want our documentation on `Mount` to state that the caller is responsible for determining whether cross-filesystem mounts are present.

For "caller == `container`" we'd want to warn the user in the case that their single file mount won't actually propagate guest writes through to the host file.

Sources/Containerization/FileMount.swift
dcantah: Both, but for the guest it's only the name in the temp holding spot. It's mostly convenience, as I could store the full path in the tempdir we make, and just use the last path component (the filename), but separating them makes it a little simpler. If you wanted to share in `/path/to/cool/file.txt`:

1. `/path/to/cool/file.txt` will end up as a hardlink at `/tmp/blah/file.txt`.
2. `/tmp/blah` gets mounted at `/run/file-mounts/some-virtiofs-tag/` in the guest
3. Edit oci config to mount `/run/file-mounts/some-virtiofs-tag/file.txt` into container at wherever they asked for it

PR #498, opened 2026-01-28
vminitd: Set nofile rlimit to max

dcantah:
No reason to default to 65k


dcantah:
Hang on, I didn't replace the init image properly. We need to set this to nr_open instead of infinity (guest is crashing)

dkovba (approved):

PR #499, opened 2026-01-28
Add LinuxRLimit type

dcantah:
I was not a huge fan of how rlimits were specified using the OCI type as it just takes in strings for the type which isn't super ergonomic. This adds a new type for rlimits that we can convert to oci type like we already do for capabilities.

### DIFF
--- a/Sources/Containerization/LinuxProcessConfiguration.swift
+++ b/Sources/Containerization/LinuxProcessConfiguration.swift
@@ -17,6 +17,193 @@
 import ContainerizationOCI
 import ContainerizationOS
 
+/// A resource limit (rlimit) configuration for a container process.
+public struct LinuxRLimit: Sendable, Hashable {
+    /// The kind of resource limit.
+    public var kind: Kind
+    /// The hard limit value.
+    public var hard: UInt64
+    /// The soft limit value.
+    public var soft: UInt64
+
+    /// Creates a new resource limit.
+    ///
+    /// - Parameters:
+    ///   - kind: The kind of resource limit.
+    ///   - hard: The hard limit value.
+    ///   - soft: The soft limit value.
+    public init(kind: Kind, hard: UInt64, soft: UInt64) {
+        self.kind = kind
+        self.hard = hard
+        self.soft = soft
+    }
+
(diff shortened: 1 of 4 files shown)


Sources/Containerization/LinuxProcessConfiguration.swift
```diff
@@ -17,6 +17,145 @@
...
+        case .openFiles:
+            "RLIMIT_NOFILE"
+        case .stackSize:
+            "RLIMIT_STACK"
+        case .numberOfProcesses:
+            "RLIMIT_NPROC"
+        case .residentSetSize:
+            "RLIMIT_RSS"
+        case .lockedMemory:
+            "RLIMIT_MEMLOCK"
```
dkovba: Should we support these resource limits as well?
- `RLIMIT_LOCKS` (Linux 2.4.0 to Linux 2.4.24)
- `RLIMIT_MSGQUEUE` (since Linux 2.6.8)
- `RLIMIT_NICE` (since Linux 2.6.12)
- `RLIMIT_RTPRIO` (since Linux 2.6.12)
- `RLIMIT_RTTIME` (since Linux 2.6.25)
- `RLIMIT_SIGPENDING` (since Linux 2.6.8)

dcantah: Yea, good catch

Sources/Containerization/LinuxProcessConfiguration.swift
```diff
@@ -17,6 +17,193 @@
...
+        case .messageQueue:
+            "RLIMIT_MSGQUEUE"
+        case .nice:
+            "RLIMIT_NICE"
+        case .realtimePriority:
+            "RLIMIT_RTPRIO"
+        case .realtimeTimeout:
+            "RLIMIT_RTTIME"
+        case .signalsPending:
+            "RLIMIT_SIGPENDING"
```
dkovba, line 200: Should we use the same order as in "getrlimit(2) — Linux manual page"?

dcantah: Good call as well, reordering

PR #500, opened 2026-01-28
Make EXT4.Inode properties public

slang25:
## Summary

The `EXT4.Inode` struct is public but its properties were internal, preventing external consumers from reading inode metadata. This change makes the properties public to enable use cases like reading file metadata (mode, uid, gid, timestamps) from ext4 filesystems.

Motivating use case: https://github.com/socktainer/socktainer/pull/169

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dcantah (approved):
Not sure why these were internal to begin with after the type itself was made public, looks good to me

PR #502, opened 2026-01-30
[keychain]: add list function + update variable/parameter names

saehejkang:
## Type of Change
- [x] Bug fix
- [x] New feature  
- [x] Breaking change
- [ ] Documentation update

## Motivation and Context

- Add a `list` function that is needed for the `container registry list` command
- Updates to `KeychainHelper` and `KeychainQuery` variable/parameter names
- Updates to test cases

## Depends On

Relates to: https://github.com/apple/container/issues/1088
Needs changes from: https://github.com/saehejkang/containerization/pull/1

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerizationOCI/Client/KeychainHelper.swift
+++ b/Sources/ContainerizationOCI/Client/KeychainHelper.swift
@@ -20,22 +20,22 @@ import ContainerizationOS
 
 /// Helper type to lookup registry related values in the macOS keychain.
 public struct KeychainHelper: Sendable {
-    private let id: String
-    public init(id: String) {
-        self.id = id
+    private let securityDomain: String
+    public init(securityDomain: String) {
+        self.securityDomain = securityDomain
     }
 
-    /// Lookup authorization data for a given registry domain.
-    public func lookup(domain: String) throws -> Authentication {
+    /// Lookup authorization data for a given registry hostname.
+    public func lookup(hostname: String) throws -> Authentication {
         let kq = KeychainQuery()
 
         do {
-            guard let fetched = try kq.get(id: self.id, host: domain) else {
+            guard let fetched = try kq.get(securityDomain: self.securityDomain, hostname: hostname) else {
                 throw Self.Error.keyNotFound
             }
             return BasicAuthentication(
-                username: fetched.account,
-                password: fetched.data
+                username: fetched.username,
+                password: fetched.password
             )
         } catch let err as KeychainQuery.Error {
             switch err {
@@ -47,38 +47,38 @@ public struct KeychainHelper: Sendable {
         }
     }
 
-    /// Lists all registry entries for this domain.
+    /// Lists all registry entries for this security domain.
     /// - Returns: An array of registry metadata for each matching entry, or an empty array if none are found.
     /// - Throws: An error if the keychain query fails.
     public func list() throws -> [RegistryInfo] {
         let kq = KeychainQuery()
-        return try kq.list(domain: self.id)
+        return try kq.list(securityDomain: self.securityDomain)
     }
 
-    /// Delete authorization data for a given domain from the keychain.
-    public func delete(domain: String) throws {
+    /// Delete authorization data for a given hostname from the keychain.
+    public func delete(hostname: String) throws {
         let kq = KeychainQuery()
-        try kq.delete(id: self.id, host: domain)
+        try kq.delete(securityDomain: self.securityDomain, hostname: hostname)
     }
 
-    /// Save authorization data for a given domain to the keychain.
-    public func save(domain: String, username: String, password: String) throws {
+    /// Save authorization data for a given hostname to the keychain.
+    public func save(hostname: String, username: String, password: String) throws {
         let kq = KeychainQuery()
-        try kq.save(id: self.id, host: domain, user: username, token: password)
+        try kq.save(securityDomain: self.securityDomain, hostname: hostname, username: username, password: password)
     }
 
-    /// Prompt for authorization data for a given domain to be saved to the keychain.
+    /// Prompt for authorization data for a given hostname to be saved to the keychain.
     /// This will cause the current terminal to enter a password prompt state where
     /// key strokes are hidden.
-    public func credentialPrompt(domain: String) throws -> Authentication {
-        let username = try userPrompt(domain: domain)
+    public func credentialPrompt(hostname: String) throws -> Authentication {
+        let username = try userPrompt(hostname: hostname)
         let password = try passwordPrompt()
         return BasicAuthentication(username: username, password: password)
     }
 
     /// Prompts the current stdin for a username entry and then returns the value.
-    public func userPrompt(domain: String) throws -> String {
-        print("Provide registry username \(domain): ", terminator: "")
+    public func userPrompt(hostname: String) throws -> String {
+        print("Provide registry username \(hostname): ", terminator: "")
         guard let username = readLine() else {
             throw Self.Error.invalidInput
         }
--- a/Sources/ContainerizationOS/Keychain/KeychainQuery.swift
+++ b/Sources/ContainerizationOS/Keychain/KeychainQuery.swift
@@ -19,8 +19,8 @@ import Foundation
 
 /// Holds the result of a query to the keychain.
 public struct KeychainQueryResult {
-    public var account: String
-    public var data: String
+    public var username: String
+    public var password: String
     public var modifiedDate: Date
     public var createdDate: Date
 }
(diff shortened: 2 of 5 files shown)


Sources/ContainerizationOCI/Client/KeychainHelper.swift
```diff
@@ -46,6 +46,12 @@ public struct KeychainHelper: Sendable {
             }
         }
     }
+    
+    /// List all registry domains this id has credentials for.
+    public func listDomains() throws -> [String] {
```
jglogan: Let's nudge this towards something where a registry can be a managed resource in container.

The caller is presumably the manager of a specific keychain domain (such as the one for container).

I think this should be a list(domain:) -> [RegistryInfo] that enumerates all keychain entries for the domain, mapping them to a new structure that has all the stored attributes for the registry except the secret value.

This structure should conform to Sendable but not Codable - the client is responsible for that. In container it will be that it eventually conforms to managed resource but could just be RegistryInfo with a Codable extension initially.

Thoughts?

Sources/ContainerizationOS/Keychain/KeychainQuery.swift
```diff
@@ -107,6 +107,35 @@ public struct KeychainQuery {
             createdDate: createdDate
         )
     }
+    
+    /// List all hostnames in the keychain.
+    public func listHosts(id: String) throws -> [String] {
```
jglogan: Do we need a public version of two calls if we implement the suggestion above?

Sources/ContainerizationOCI/Client/KeychainHelper.swift
saehejkang: > I think this should be a list(domain:) -> [RegistryInfo] that enumerates all keychain entries for the domain, mapping them to a new structure that has all the stored attributes for the registry except the secret value.

The `domain` that is passed into the `list` function comes from the manager of the specific keychain correct?

________

> This structure should conform to Sendable but not Codable 

This would be the `RegistryInfo` model?
________

Is this how the relationship between the domain/registry sorta looks? It is not 1:1 of what exists now, but it might help me visualize what we want to accomplish here.

```
containerDomain1
 ├─ Keychain #1
 │    registry: docker.io
 │    username: alice
 │    secret: ********
 │
 ├─ Keychain #2
 │    registry: ghcr.io
 │    username: alice
 │    secret: ********
```

Sources/ContainerizationOS/Keychain/KeychainQuery.swift
saehejkang: I believe we do not

Sources/ContainerizationOCI/Client/KeychainHelper.swift
jglogan: Let's look at a corner case of a private registry on a non-standard port.

```
Application.Login.run()
    let keychain = KeychainHelper(id: "com.apple.container.registry")
    // successful login
    try keychain.save(domain: "registry.isis.com:1234", username: "mallory", password: "guest")
        let kq = KeychainQuery()
        try kq.save(id: "com.apple.container.registry", host: "registry.isis.com:1234", user: "mallory", token: "guest")
            if try exists(id: "com.apple.container.registry", host: "registry.isis.com:1234") {
                try delete(id: "com.apple.container.registry", host: "registry.isis.com:1234")
            }
            let query: [String: Any] = [
                kSecClass as String: kSecClassInternetPassword,
                kSecAttrSecurityDomain as String: "com.apple.container.registry",
                kSecAttrServer as String: "registry.isis.com:1234",
                kSecAttrAccount as String: "mallory",
                kSecValueData as String: tokenEncoded,
                kSecAttrAccessible as String: kSecAttrAccessibleAfterFirstUnlock,
                kSecAttrSynchronizable as String: false,
            ]
            let status = SecItemAdd(query as CFDictionary, nil)
```

The parameter names on KeychainHelper feel a little janky.

The initialization function would be better as `init(securityDomain:)`.

The save function: `save(host:username:password)`.

To your questions:

`container` manages exactly one keychain security domain, named `com.apple.container.registry`. Other applications built over the containerization library would define their own security domains.

RegistryInfo should be Sendable but not Codable. We need to merge apple/container#1123 for Monday's release so plan to rebase after that, and then in container you can define a RegistryResource that conforms to ManagedResource, and transform what comes back from KeychainHelper to objects of that type.

I would expect the list function to return a RegistryInfo object for each internet password in the `com.apple.container.registry` keychain security domain. The invariant is that there is an internet password for a hostname if and only if the user has performed a valid `registry login {hostname}` and has not subsequently performed `registry logout {hostname}`.

So if you change the parameters on all the KeychainHelper funcs to be more clear, does all of that make sense?

saehejkang: > does all of that make sense

Makes sense to me! I will come back to making updates to all of the functions/parameters in the `KeychainHelper`. I want to keep the scope of this PR to just listing out the registries. I’ll push these updates shortly for you to take a look, but I won’t mark the PR ready for final review until I’ve added test cases.

Sources/ContainerizationOCI/Client/KeychainHelper.swift
```diff
@@ -47,6 +47,12 @@ public struct KeychainHelper: Sendable {
         }
     }
 
+    /// List all registry entries for this domain.
```
jglogan: Could you make the following formal parameter changes in KeychainHelper and KeychainQuery?

- Use `username` and `password` for account name and credentials (not `account` or `user`).
- Use `securityDomain` instead of `id`.
- Use `hostname` instead of `domain` or `host` (or `registry domain` in docc).

Also can you full docc (description and parameters) on both types?

Sources/ContainerizationOS/Keychain/KeychainQuery.swift
```diff
@@ -25,6 +25,14 @@ public struct KeychainQueryResult {
     public var createdDate: Date
 }
 
+/// Holds the stored attributes for a registry.
```
jglogan: Put this in its own file, and add docc for the fields (hostname is a domain name and optional port, for example).

Sources/ContainerizationOCI/Client/KeychainHelper.swift
saehejkang: https://github.com/saehejkang/containerization/pull/1 for updates to params that will just be merged into this PR

jglogan: What do you think? When I look at that PR it reads much more clearly to me.

jglogan: There's another PR that changes the API slightly so once we can get the 0.9.0 release out we can merge this fix and the other one, bump the API to 0.25.0, and tackle the corresponding changes on the container side.

saehejkang: Much clearer to me. I had to open the separate PR just cuz I was mixing up what was what and wanted to make sure I didn't miss anything lol

saehejkang: I am just about to push up changes for the `container` side right now (will mark as draft)

jglogan:
@saehejkang Were you going to merge the parameter changes you made in the draft?

saehejkang:
> Were you going to merge the parameter changes you made in the draft?

Here is a checklist of everything that needs to be done

- [x] I need a review of https://github.com/saehejkang/containerization/pull/1. Just want to make sure it looks correct and once I get the go ahead, I will merge it into this PR.
- [x] I need a review/merge of this PR (`containerization/502`)
- [x] I will go back to https://github.com/apple/container/pull/1138 and make sure it is up to date and working locally
- [x] I then need a review of `container/1138` PR, with it ready to be merged
- [x] Bump `containerization` to `0.25.0`, as well as in the `container` project
- [x] Finally, merge the `container/1138` PR

Feels like a long list of demands 😅 . I believe this is the correct workflow, as I don’t want things to be merged or bumped out of order. There are quite a few breaking changes, so I wanted to lay everything out step by step.

jglogan:
@saehejkang Appreciate the thoughtful approach, I glanced at the PR and approved because I'd gone over it closely when I first saw it. I could have sworn I commented then (along the lines of "this looks way clearer, don't you think?") but now I can't find it.

saehejkang:
> I could have sworn I commented then (along the lines of "this looks way clearer, don't you think?") but now I can't find it.

you did but I just wanted to be sure with your ✅ of approval 😆

jglogan (approved):

jglogan:
@saehejkang merged!

PR #507, opened 2026-02-01
Fix usage of "clean up" as verb

DePasqualeOrg:
"clean up" is a verb, and "cleanup" is a noun. There are several examples of the incorrect usage of "cleanup" as a verb in this repository, including one in the public API (`cleanupOrphanedBlobs`). For this public method, I've added a deprecation notice to the name with the incorrect usage.

### DIFF
--- a/Sources/Containerization/Image/ImageStore/ImageStore.swift
+++ b/Sources/Containerization/Image/ImageStore/ImageStore.swift
@@ -141,12 +141,6 @@ extension ImageStore {
         }
     }
 
-    @available(*, deprecated, renamed: "cleanUpOrphanedBlobs")
-    @discardableResult
-    public func cleanupOrphanedBlobs() async throws -> (deleted: [String], freed: UInt64) {
-        try await cleanUpOrphanedBlobs()
-    }
-
     /// Calculate the size of orphaned blobs without deleting them.
     ///
     /// - Returns: The total size in bytes of blobs that are not referenced by any image.


Sources/Containerization/Image/ImageStore/ImageStore.swift
```diff
@@ -135,12 +135,18 @@ extension ImageStore {
...
     @discardableResult
-    public func cleanupOrphanedBlobs() async throws -> (deleted: [String], freed: UInt64) {
+    public func cleanUpOrphanedBlobs() async throws -> (deleted: [String], freed: UInt64) {
         try await self.lock.withLock { lockCtx in
-            try await self._cleanupOrphanedBlobs(lockCtx)
+            try await self._cleanUpOrphanedBlobs(lockCtx)
         }
     }
 
+    @available(*, deprecated, renamed: "cleanUpOrphanedBlobs")
```
jglogan: Could you remove the old function entirely, and update any call sites in containerization with the renamed function name? We're not at the point in this project where we're concerned about breaking the API (we will be getting there soon).

jglogan (changes requested):

Sources/Containerization/Image/ImageStore/ImageStore.swift
DePasqualeOrg: Okay, I've done that.

jglogan: Thanks! Building now.

jglogan:
@DePasqualeOrg could you rebase your change onto current main and push again? That way you'll have #511 which will help ensure tests pass (not the fault of your changes).

DePasqualeOrg:
I've rebased on main.

jglogan:
@DePasqualeOrg Thanks for the contribution, I'd like to get the registry API changes from @saehejkang in and then we'll bump the tag and then you can update your container PR to pick up the updated API.

PR #514, opened 2026-02-04
[github]: add pull request template

saehejkang:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

- Adds a pull request template

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/bug_fix_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/bug_fix_template.md
@@ -1,0 +1,27 @@
+### Bug Description
+<!-- Describe the bug that was resolved with this PR. -->
+
+### Resolution Description
+<!-- Describe how the bug was resolved with this PR. -->
+
+### Related Issue (if applicable)
+<!-- Link to the issue that reported this bug. -->
+
+### Testing Details (if applicable)
+<!-- Add details on how to appropriately test changes made in this PR. -->
+
+### Review Checklist
+<!-- This checklist should be completed before the PR is eligible for review. -->
+
+- [ ] The bug fix fully resolves the issue.
+- [ ] Tests have been added or updated appropriately.
+- [ ] Documentation has been updated appropriately.
+- [ ] Commented-out code and debugging artifacts have been removed.
+- [ ] Commit messages are clear and appropriate.
+- [ ] The PR title accurately reflects the code changes.
+
+### Code of Conduct
+<!-- By submitting this PR, you agree to follow the project's Code of Conduct. -->
+
+- [ ] I agree to follow this project's [Code of Conduct](https://github.com/apple/.github/blob/main/CODE_OF_CONDUCT.md)
+
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,3 @@
+Go to the `Preview` tab and select the appropriate template, based on your changes:
+
+* [Bug Fix](?expand=1&template=bug_fix_template.md)


jglogan:
@saehejkang lol I kind of want to get rid of the one over in container 😄 

Or refine what's there. Let's keep this open for now and I'll take it up with the other maintainers today.

katiewasnothere:
I agree with @jglogan :/ 

I feel like the only reason I'm okay with the template in `container` is that that's where we get most of our contributions and it's nice to remind people to write good PR descriptions and test things. I *really* wish GitHub had PR templates that were similar to issue templates so we could have more complex fields. 

My main issues with the PR templates as they are today are: 
1. We use squashed commits which then uses the PR description as the final commit message once the commits are squashed. The template makes those commits probably less clear than if we just kept the commit messages as they were from the PR. But the issue with that is that not everyone knows how to write good commit messages and we don't want to be staling PRs on the author needing to just change the commit messages. 
2. We can't add conditional fields. It'd be nice if we could have something like if someone checks that they added a new feature that they're REQUIRED to also check that they have added tests. 
3. We can't add mandatory fields. I like that in the issue templates we can require that people agree to the code of conduct for the project. 
4. The GitHub UI treats the checkboxes in a PR description as like "todo" task items, so when you're looking at the list of PRs (https://github.com/apple/container/pulls), they'll say something like "1 of 7 tasks", which is unclear and annoying.

jglogan:
@katiewasnothere Re #1, I wind up editing all commit messages that I merge for conciseness (except where I forget to).

My opinion is that it's okay for maintainers to have a bit of editorial control on the commit messages. We're not mucking with the content of the PRs and it feels nitpicky and inefficient for us to go back and forth with contributors over the messages.

The task lists convey very little info to me. Would it work to put our "PR submission reminders" (write tests! write documentation!) `<!-- in comments? -->` I think they may appear when opening the PR but will get filtered for display.

katiewasnothere:
@jglogan Yeah I definitely think maintainers can have some editorial control on the final commit message. It's more just annoying that we need to edit the final commit message and we may forget. 

I would prefer we not put the reminders in a comment in the template. I like that the checkbox guides people to explicitly confirm they've done something like add tests, etc.

saehejkang:
I have read all the comments and may have an idea on a solution that solves all these issues 👀 . For a first pass, I will only focus on two types of PRs: a bug fix or new feature. In a way, this solution could also provide a new way to determine when a PR is FULLY COMPLETE before review, since we want the checkbox guides to be completed. Let me push something up here in the next couple of days and we can continue the discussion then.

saehejkang:
@jglogan @katiewasnothere I pushed changes for a bug fix template. The other templates will be very similar, so it warrants a review before I continue. I hope this is sort of in the right direction, as it allows template tailoring for different types of PRs (bugs, features, refactoring, etc.)

katiewasnothere:
@saehejkang My understanding is that to use these, the PR submitter needs to add the path to the template in the url when they go to make a PR. Is there another way of using these?

saehejkang:
Apologies! I forgot the most important part of the changes and missed some things 😅 . 

>  Is there another way of using these?

The `pull_request_template.md` file should autofill any new PR descriptions and devs can simply route to the `Preview` tab and click on a linked template. That then will autofill the description with the new template and there is no need to add the path to the template in the url.

However, the only way to test these changes would be to merge it to `main` and then open a test PR.

katiewasnothere:
@saehejkang interesting. I don't know how I feel about having people click on a template. I feel like people may just choose to remove the full text and put their own information in it without following any template? That said, I do like having something people can follow and that we can point people to. 

@jglogan What do you think?

saehejkang:
> I feel like people may just choose to remove the full text and put their own information in it without following any template?

That makes sense, but having a proper structure should help with adoption. Even with the current template, devs do generally try to follow it as best they can. I haven't seen anyone really go out of their way to remove it and its entirety because they did not like it 😆 .

dcantah:
To me, and this is ridiculously subjective obviously, I wouldn't want a template here. The volume of contributions here is somewhat sane, and I don't find much value in the template (and this harkens back to what John said above that it could possibly be better, then maybe my opinion would change) for container.

saehejkang:
> The volume of contributions here is somewhat sane, and I don't find much value in the template

If the maintainer sentiment is to not include a template in the `containerization` repo, I will update this PR to remove the original template completely. 

Furthermore, the volume of contributions is much higher in the `container` repo, so updating the template there is more warranted. We should just continue the discussion here before opening up another PR there.

dcantah:
Anyone feel free to reopen if they find value, but we can revisit this later if contribution volume increases significantly

saehejkang:
@dcantah Reopen this PR to maybe just strip out the template entirely?

dcantah:
Do we have a template here? On mobile, but I thought this PR was adding one because we didn't

saehejkang:
> Do we have a template here?

Yes, there is a template.

dcantah:
We have issue templates, which I think are good, but we have a pr template? Am I blind 😭

saehejkang:
> Am I blind

Nope I am blind. I got mixed up because I copied the template from `container` for this PR so I assumed there was one 😭 .

PR #522, opened 2026-02-07
LinuxContainer/LinuxPod: Support duplicate virtiofs mounts

dcantah:
If we have the same source we should dedupe these. Previously virt.framework would yell at us for a duplicate tag because we derive the tag from hashing the mount source.


dcantah:
Fixes https://github.com/apple/container/issues/380

PR #529, opened 2026-02-10
Add a public method to wait for the relay to complete

dkovba:
- Adds a public method to wait for the relay to complete when both directions close
- Improves error handling
- Moves `UnixSocketRelayManager` into a separate file
- Renames `SocketRelay` to `UnixSocketRelay`
- Adds logs at the info level after converting other logs to the debug level previously

### DIFF
--- a/vminitd/Sources/vminitd/Server+GRPC.swift
+++ b/vminitd/Sources/vminitd/Server+GRPC.swift
@@ -211,6 +211,7 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
             try await proxy.start()
             try await state.add(proxy: proxy)
         } catch {
+            try? await proxy.close()
             log.error(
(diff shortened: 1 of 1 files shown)


vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -198,16 +198,16 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
...
+        let proxy = VsockProxy(
+            id: request.id,
+            action: request.action == .into ? .dial : .listen,
+            port: request.vsockPort,
+            path: URL(fileURLWithPath: request.guestPath),
+            udsPerms: request.guestSocketPermissions,
+            log: log
+        )
 
+        do {
```
adityaramani, line 210: Add a `try? proxy.stop()` in the catch block?

Issue #494, opened 2026-01-26
[Request]: Allow vminitd log level to be configured.


jglogan:
### Feature or enhancement request details

When I bootstrap a VM the default log level should be info, and it should be configurable to that I can see debug and trace messages as needed.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

manojmahapatra:
If this has not been started, I would like to take a stab at it. @jglogan do you have anything specific in mind? If not, I have some questions;

1. Should we support only `info`, `debug`, `trace`, or other Swift log levels as well?
2. If user passes invalid level (e.g., `--log-level foo`), should we:
   - default to `info` silently
   - log a warning and default to `info`
   - fail with an error
3. What level of testing is expected? Is a unit test for arg parsing sufficient?

PR #531, opened 2026-02-13
vminitd: Add ability to set log level

dcantah:
This adds a small flag named --log-level to be able to set the log level of vminitd and the pause command. I imagine this won't be the last time we'll have to add flags to vminitd so I went ahead and converted vminitd to be a Argument Parser based binary as well.

### DIFF
--- a/vminitd/Sources/vminitd/Application.swift
+++ b/vminitd/Sources/vminitd/Application.swift
@@ -14,45 +14,43 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
+import ArgumentParser
 import ContainerizationOS
 import Foundation
 import Logging
 
 @main
-struct Application {
+struct Application: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "vminitd",
+        abstract: "Virtual machine init daemon",
+        version: "0.1.0",
+        subcommands: [
+            InitCommand.self,
+            PauseCommand.self,
+        ],
+        defaultSubcommand: InitCommand.self
+    )
+
     static func main() async throws {
-        LoggingSystem.bootstrap(StreamLogHandler.standardError)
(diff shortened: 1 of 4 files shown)


vminitd/Sources/vminitd/Application.swift
```diff
@@ -14,46 +14,44 @@
...
-        LoggingSystem.bootstrap(StreamLogHandler.standardError)
-
-        // Parse command line arguments
-        let args = CommandLine.arguments
-        let command = args.count > 1 ? args[1] : "init"
+struct Application: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "vminitd",
+        abstract: "Virtual machine init daemon",
+        version: "0.1.0",
```
katiewasnothere, line 27: Should this be the version of the previous tag + commit info if not at a tagged commit?

vminitd/Sources/vminitd/Application.swift
```diff
@@ -14,46 +14,44 @@
...
+        // Swift has issues spawning threads if /proc isn't mounted,
+        // so we do this synchronously before any async code runs.
+        try mountProc()
 
-            log.info("Running init command")
-            try Self.mountProc(log: log)
-            try await InitCommand.run(log: log)
+        var command = try parseAsRoot()
+        if let asyncCommand = command as? AsyncParsableCommand {
+            nonisolated(unsafe) var unsafeCommand = asyncCommand
```
katiewasnothere, line 42: why does this need to be an unsafe variable?

vminitd/Sources/vminitd/Application.swift
dcantah: We could, I'm not too worried about the version at the moment

dcantah:
https://github.com/apple/containerization/issues/494

PR #533, opened 2026-02-18
LinuxContainer: Add support for optional writable layer

dcantah:
This adds support to LinuxContainer to be able to provide an optional writable layer. This is useful as today the rootfs size is determined by the size of the image block itself which is set in stone at unpack time of the image. This leaves a lot to be desired because a user might want a larger or smaller rootfs size, but it's not configurable on a per container basis because of this. To support this, we can pass an additional writable block device that we can overlayfs with the image contents in the guest. All writes will go to this writable layer, and the rootfs size in the container is now whatever size this writable layer is.

The block devices (via our ext4 package, but you could use whatever) are cheap enough to generate on the fly that you can choose whatever size you want easily.


Sources/Containerization/LinuxContainer.swift
```diff
@@ -313,7 +342,8 @@ public final class LinuxContainer: Container, Sendable {
 
         // If the rootfs was requested as read-only, set it in the OCI spec.
         // We let the OCI runtime remount as ro, instead of doing it originally.
-        spec.root?.readonly = self.rootfs.options.contains("ro")
+        // However, if we have a writable layer, the overlay allows writes so we don't mark it read-only.
+        spec.root?.readonly = self.rootfs.options.contains("ro") && self.writableLayer == nil
```
jglogan, line 346: Or:

```swift
guard !self.rootfs.options.contains("ro") || self.writableLayer == nil else {
    throw ContainerizationError(.invalidArgument, message: "cannot define writable layer with read-only root fs")
}

spec.root?.readonly = self.rootfs.options.contains("ro")
```

jglogan: If we're going to rework the root mount stuff whatever we do here will change. Either approach is correct.

PR #534, opened 2026-02-18
fix(dns): validate nameserver addresses are valid IP addresses

### LINKED ISSUE #467: [Request]: IPv6: validate nameserver addresses in DNSConfiguration

Right now any string can be used as a nameserver. We should consider admitting only IP addresses.

### Code of Conduct

cluster2600:
## What

Add a `DNS.validate()` method that verifies all nameserver entries are valid IPv4 or IPv6 addresses. The method is called from `Vminitd.configureDNS()` before applying the configuration.

## Why

Closes #467. Currently, any arbitrary string can be passed as a nameserver in `DNSConfiguration`, which silently results in an invalid `/etc/resolv.conf` inside the container. Hostname strings like `dns.example.com` would be written to resolv.conf but would not work as nameservers.

## How

- Added `DNS.validate() throws` method in `DNSConfiguration.swift` that iterates over all nameservers and attempts to parse each as either an `IPv4Address` or `IPv6Address` (using the existing parsers in `ContainerizationExtras`)
- Added `import ContainerizationExtras` to `DNSConfiguration.swift` (the `Containerization` target already depends on `ContainerizationExtras`)
- Called `config.validate()` at the start of `Vminitd.configureDNS()` so validation happens before any GRPC call

## Testing

Added 6 new unit tests in `DNSTests.swift`:
- ✅ Valid IPv4 nameservers accepted
- ✅ Valid IPv6 nameservers accepted  
- ✅ Mixed IPv4/IPv6 nameservers accepted
- ✅ Empty nameserver list accepted
- ❌ Hostname rejected
- ❌ Invalid address rejected

## Checklist

- [x] Tests added
- [x] No breaking changes to existing API (validate() is a new method, existing init is unchanged)
- [x] Follows existing patterns (uses ContainerizationExtras IP address types, ContainerizationError for errors)

### DIFF
--- a/Sources/Containerization/DNSConfiguration.swift
+++ b/Sources/Containerization/DNSConfiguration.swift
@@ -14,6 +14,9 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
+import ContainerizationError
+import ContainerizationExtras
+
 /// DNS configuration for a container. The values will be used to
 /// construct /etc/resolv.conf for a given container.
 public struct DNS: Sendable {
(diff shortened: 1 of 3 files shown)


dcantah:
@cluster2600 Howdy, you'll need to sign your commit (just squash to one commit please), but this change looks good to me so far.

cluster2600:
@dcantah Thanks for the review! Just squahsed everything into a single signed commit 👍

cluster2600:
thank you all

PR #541, opened 2026-02-19
ContainerizationOCI: raise Reference name limit to 255 to match OCI spec

### LINKED ISSUE #453: [Request]: Allow repository length to be greater than 127 symbols

There is validation on the length of the repository which limits you to 127 symbols wether you try to run or build with a container image.

For example:
xxxxxxx-xxxxxxxxx-xxxxxxxx.xxx.xxxxxxxx.xxx/xxxxxx-xxxxxx/xxxxxx-xxx.xx-xxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxx/xxxxxxx/xxxxx/xxxxxx/xxxxxxx-xxxxx/xxxxxx:vv-vvvvv 
Triggers:  Error: invalidArgument: "Repo length 170 greater than 127"   
Where 170 in this case is the length of the repository without the tag

This validation is consistent from from 0.1.0 until current  0.7.1

### Code of Conduct

cluster2600:
## What

Raise the  constant from 127 to 255 in `Reference.swift` and update the derived `referenceTotalLengthMax` accordingly.

## Why

Closes #453.

The [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md) states that the **name** component of an image reference (registry host + repository path) may be at most 255 bytes. The previous hard-coded limit of 127 characters incorrectly rejected valid references that used long registry hostnames or deep path hierarchies.

The old `referenceTotalLengthMax = 255` was also inconsistent: with a 127-char name cap, a 255-char total reference would only allow a very short tag. The new value is derived explicitly as `nameTotalLengthMax (255) + separator (1) + tagLengthMax (128) = 384`.

## How

- `nameTotalLengthMax`: 127 → 255
- `tagLengthMax`: new constant (128) documenting the maximum tag length already enforced by the tag regex (`{0,127}` + leading char = 128 chars).
- `referenceTotalLengthMax`: computed from the two constants above (384) rather than hard-coded to 255.

## Testing

Added three new cases in `ReferenceTests.swift`:
- ✅ Name of 128 characters (registry + path) — previously rejected, now accepted
- ✅ Name of exactly 255 characters — at the OCI spec maximum, accepted
- ❌ Name of 256 characters — one over the limit, rejected

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/apple/containerization/blob/main/CONTRIBUTING.md)
- [x] My changes follow the existing code style
- [x] I have added tests that cover the fix
- [x] All commits are GPG-signed

### DIFF
--- a/Sources/ContainerizationOCI/Reference.swift
+++ b/Sources/ContainerizationOCI/Reference.swift
@@ -17,8 +17,13 @@
 import ContainerizationError
 import Foundation
 
-private let referenceTotalLengthMax = 255
-private let nameTotalLengthMax = 127
+// nameTotalLengthMax matches the OCI distribution spec which allows up to 255 bytes for the
+// repository name component (domain + "/" + path).
+private let nameTotalLengthMax = 255
+// referenceTotalLengthMax is the upper bound for the full reference string: max name (255) +
+// separator (1) + max tag length (128) = 384.
+private let tagLengthMax = 128
+private let referenceTotalLengthMax = nameTotalLengthMax + 1 + tagLengthMax
 private let legacyDockerRegistryHost = "docker.io"
 private let dockerRegistryHost = "registry-1.docker.io"
 private let defaultDockerRegistryRepo = "library"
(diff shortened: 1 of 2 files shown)


jglogan:
@cluster2600 Thank you for the submission!

@adityaramani Assigned you to review when you have a moment. I pasted some code from an earlier implementation to issue #453 for comparison.

PR #545, opened 2026-02-20
Downgrade a relay error to a warning

dkovba:
This PR downgrades a relay error that might be logged when a connection is closed to a warning. Related PR in container: https://github.com/apple/container/pull/1238

### DIFF
--- a/Sources/ContainerizationOS/Socket/BidirectionalRelay.swift
+++ b/Sources/ContainerizationOS/Socket/BidirectionalRelay.swift
@@ -211,7 +211,14 @@ public final class BidirectionalRelay: Sendable {
                 to: destinationFd
             )
         } catch {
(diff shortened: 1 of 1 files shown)


Sources/ContainerizationOS/Socket/BidirectionalRelay.swift
```diff
@@ -211,7 +211,7 @@ public final class BidirectionalRelay: Sendable {
                 to: destinationFd
             )
         } catch {
-            log?.error("file descriptor copy failed \(error)")
+            log?.warning("file descriptor copy failed \(error)")
```
jglogan: we should be using structured logging here

dkovba: Done

jglogan (approved):

PR #559, opened 2026-02-25
Mount: Ensure symlinks stay scoped to rootfs

dcantah:
Today we don't resolve symlinks relative to the rootfs of the container as ContainerizationOS.Mount did not handle this. This means trying to mount something at /var/run (which typically is a symlink to /run in most cases) would resolve to /run in the root of the VM. This change adds some logic to handle this.

```
➜  ./bin/cctl run --kernel bin/kernel.arm64 --image docker.io/library/ubuntu:latest --mount /Users/dcantah/foo:/var/run/foo bash
root@sandbox-vm:/# ls var/run/foo/
howdy
```

### DIFF
--- a/Sources/ContainerizationOS/Mount/Mount.swift
+++ b/Sources/ContainerizationOS/Mount/Mount.swift
@@ -14,6 +14,7 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
+import CShim
 import Foundation
 
 #if canImport(Musl)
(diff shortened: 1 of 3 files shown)


Sources/ContainerizationOS/Mount/Mount.swift
```diff
@@ -116,13 +117,166 @@ extension Mount {
...
-        try self.mountToTarget(target: rootURL.path, createWithPerms: createWithPerms)
+        let resolvedTarget = try secureResolveInRoot(root: root)
+        defer { close(resolvedTarget.fd) }
+
+        let realPath = try readlinkProc(fd: resolvedTarget.fd)
+        try self.mountToTarget(target: realPath, createWithPerms: createWithPerms, targetResolved: true)
+    }
+
+    /// Result of securely resolving a path within a root directory.
+    private struct SecureResolveResult {
```
yibozhuang: Just for my own understanding: is there a reason to create this struct? Is it just for readability?

Sources/ContainerizationOS/Mount/Mount.swift
```diff
@@ -116,13 +117,166 @@ extension Mount {
...
+                    close(rootFd)
+                    throw Error.errno(savedErrno, "failed to re-open mount point file '\(component)'")
+                }
+                if currentFd != rootFd { close(currentFd) }
+                close(rootFd)
+                return SecureResolveResult(fd: pathFd)
+            }
+
+            guard mkdirat(currentFd, component, 0o755) == 0 else {
+                let savedErrno = errno
```
yibozhuang: Could we refactor these lines into a function nested here? Can be error prune having to repeat this multiple times here and below

Sources/ContainerizationOS/Mount/Mount.swift
dcantah: No, truthfully. It used to hold a path to /proc/{pid}/fd/{N} for a fallback approach, but I removed the fallback approach that doesn't use openat's RESOLVE_IN_ROOT

dcantah: So now it's just a dumb FD holder. I can just pass the fd around directly if you'd prefer (I'd honestly prefer as well..)

Sources/ContainerizationOS/Mount/Mount.swift
dcantah: Completely fair, let me see

Sources/ContainerizationOS/Mount/Mount.swift
yibozhuang: I think just pass `fd` would probably be better

PR #560, opened 2026-02-25
Fall back to the referrers tag schema when the referrers API is not available

realrajaryan:
Fall back to the referrers tag schema when the referrers API is not available


Sources/ContainerizationOCI/Client/RegistryClient+Referrers.swift
```diff
@@ -54,5 +56,36 @@ extension RegistryClient {
...
         }
+
+        return result
+    }
+
+    /// Fallback for registries that don't support the referrers API.
+    ///
+    /// Uses the OCI referrers tag schema: referrers for a digest are stored as an
+    /// index at the tag `<algorithm>-<hex>` (e.g., `sha256-abc123...`).
+    private func referrersTagFallback(name: String, digest: String, artifactType: String? = nil) async throws -> Index {
```
dcantah: Doesn't seem like this can actually throw

dcantah (approved):

PR #561, opened 2026-02-26
LinuxPod: Add hostname field on pod config

dcantah:
Just like we have for hosts and dns configurations, we can add a hostname field to make it easier to setup if every container is going to be using the same hostname.


Sources/Containerization/LinuxPod.swift
```diff
@@ -51,6 +51,9 @@ public final class LinuxPod: Sendable {
         /// Whether containers in the pod should share a PID namespace.
         /// When enabled, all containers can see each other's processes.
         public var shareProcessNamespace: Bool = false
+        /// The default hostname for all containers in the pod.
+        /// Individual containers can override this by setting their own `hostname` configuration.
+        public var hostname: String?
```
yibozhuang, line 56: Should we be doing
```swift
public var hostname: String = ""
```

at the pod configuration level and then at container level it would be
```swift
public var hostname: String?
```
to allow container to optionally override, this would make it consistent with the other fields

Are we ok with defaulting the hostname to "" ?

dcantah: I honestly just think we should move these all to optionals.

PR #562, opened 2026-02-27
linux pod: exec process should inherit container's process config

yibozhuang:
This change fixes an issue in linux pod where the exec process
does not inherit the container process config which means any
environment variables, working directory, user, rlimits,
capabilities would be different for the exec process.

This change ensures that exec process would inherit
environment variables, working directory, user, rlimits,
and capabilities from the container process config but
resets arguments, terminal, and stdio as those should not
be inherited.

This change also adds an integration test to exercise this.

### DIFF
--- a/Sources/Containerization/LinuxPod.swift
+++ b/Sources/Containerization/LinuxPod.swift
@@ -778,7 +778,14 @@ extension LinuxPod {
             }
 
             var spec = self.generateRuntimeSpec(containerID: containerID, config: container.config, rootfs: container.rootfs)
-            var config = LinuxProcessConfiguration()
+            // Inherit environment variables, working directory, user, capabilities, rlimits from container process.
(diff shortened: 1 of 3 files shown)


Sources/Containerization/LinuxPod.swift
```diff
@@ -778,7 +778,7 @@ extension LinuxPod {
             }
 
             var spec = self.generateRuntimeSpec(containerID: containerID, config: container.config, rootfs: container.rootfs)
-            var config = LinuxProcessConfiguration()
+            var config = container.config.process
```
dkovba, line 783: Should we reset these properties?
```
config.arguments = []
config.terminal = false
config.stdin = nil
config.stdout = nil
config.stderr = nil
```

yibozhuang: Thanks! Fixed.

dcantah (approved):

Issue #543, opened 2026-02-19
[Request]: Add directory copying support


simone-panico:
### Feature or enhancement request details

[apple/containerization#463](https://github.com/apple/containerization/pull/463) added a streaming RPC to copy single files in and out of a running container. As noted there, directory copying was intentionally left out pending a solution for tar/untar capabilities in the guest agent.

This issue tracks extending the agent to support recursive directory copying, as a follow-up to the single-file implementation.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

simone-panico:
The Feature was added in [containerization#571](https://github.com/apple/containerization/pull/571) :)

PR #563, opened 2026-02-27
feat: Copy Directory

simone-panico:
This pull request introduces functionality to support copying directories between the host and container [#543](https://github.com/apple/containerization/issues/543).

See [#1190](https://github.com/apple/container/pull/1190) for more context.

### DIFF
--- a/Sources/Containerization/Vminitd.swift
+++ b/Sources/Containerization/Vminitd.swift
@@ -653,32 +653,48 @@ extension Vminitd {
     /// Extract a tar archive to a destination directory.
     private static func extractTar(reader: TarReader, to destURL: URL) throws {
         let fm = FileManager.default
+        try fm.createDirectory(atPath: destURL.path, withIntermediateDirectories: true)
+        let fd = try FileDescriptor.open(FilePath(destURL.path), .readOnly)
+        defer { try? fd.close() }
+
         while let header = try reader.nextHeader() {
-            let relativePath = header.path.trimmingCharacters(in: CharacterSet(charactersIn: "/"))
-            guard !relativePath.contains("..") else {
+            let memberPath = FilePath(header.path.trimmingCharacters(in: CharacterSet(charactersIn: "/")))
+            guard let lastComponent = memberPath.lastComponent else {
                 try reader.skipRemainingContent()
                 continue
             }
-            let fullURL = relativePath.isEmpty ? destURL : destURL.appending(path: relativePath)
+            let relativePath = memberPath.removingLastComponent()
 
-            switch header.entryType {
-            case .directory:
-                try fm.createDirectory(at: fullURL, withIntermediateDirectories: true)
-            case .regular, .regularAlt, .contiguous:
-                let parentDir = fullURL.deletingLastPathComponent()
-                try fm.createDirectory(at: parentDir, withIntermediateDirectories: true)
-                let fd = open(fullURL.path, O_WRONLY | O_CREAT | O_TRUNC, mode_t(header.mode > 0 ? header.mode : 0o644))
-                guard fd != -1 else {
+            do {
+                switch header.entryType {
+                case .directory:
+                    try fd.mkdirSecure(memberPath, makeIntermediates: true)
+                case .regular, .regularAlt, .contiguous:
+                    try fd.mkdirSecure(relativePath, makeIntermediates: true) { dirFd in
+                        try? dirFd.unlinkRecursiveSecure(filename: lastComponent)
+
+                        let maskedMode = mode_t(header.mode & 0o777)
+                        let fileMode = maskedMode > 0 ? maskedMode : mode_t(0o644)
+                        let fileFd = openat(dirFd.rawValue, lastComponent.string, O_WRONLY | O_CREAT | O_EXCL | O_NOFOLLOW, fileMode)
+                        guard fileFd >= 0 else {
+                            try reader.skipRemainingContent()
+                            return
+                        }
+                        let fileDescriptor = FileDescriptor(rawValue: fileFd)
+                        defer { try? fileDescriptor.close() }
+                        try reader.readFile(to: fileDescriptor)
+                    }
+                case .symbolicLink:
+                    try fd.mkdirSecure(relativePath, makeIntermediates: true) { dirFd in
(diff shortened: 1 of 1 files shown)


dcantah:
I don't think we need a new set of rpcs. Presumably we can reuse the copyIn/copyOut and if you need extra info just add it to the proto message

jglogan:
@simone-panico @dcantah One thing to keep an eye on is when this lands in the swift SDK: https://github.com/swiftlang/swift-docker/commit/ea31a3331dd0a3346ae7cc114daffe4f6be35415.

At that point we might be able to replace any homegrown archiver we use with libarchive.

simone-panico:
> libarchive

So that means [ContainerizationTar](https://github.com/apple/containerization/pull/474) is not sure to be merged?

Sources/Containerization/Vminitd.swift
```diff
@@ -528,6 +530,159 @@ extension Vminitd {
...
+                guard fd != -1 else {
+                    try reader.skipRemainingContent()
+                    continue
+                }
+                let fileFD = FileDescriptor(rawValue: fd)
+                defer { try? fileFD.close() }
+                try reader.readFile(to: fileFD)
+            case .symbolicLink:
+                try? fm.removeItem(at: fullURL)
+                try fm.createSymbolicLink(atPath: fullURL.path, withDestinationPath: header.linkName)
```
jglogan: This allows the creation of symlinks that refer anywhere, which can be written through by a subsequent regular file write.

Sources/Containerization/Vminitd.swift
```diff
@@ -528,6 +530,159 @@ extension Vminitd {
...
+                continue
+            }
+            let fullURL = relativePath.isEmpty ? destURL : destURL.appending(path: relativePath)
+
+            switch header.entryType {
+            case .directory:
+                try fm.createDirectory(at: fullURL, withIntermediateDirectories: true)
+            case .regular, .regularAlt, .contiguous:
+                let parentDir = fullURL.deletingLastPathComponent()
+                try fm.createDirectory(at: parentDir, withIntermediateDirectories: true)
```
jglogan: Susceptible to TOCTOU attack.

jglogan:
> > libarchive
> 
> So that means [ContainerizationTar](https://github.com/apple/containerization/pull/474) is not sure to be merged?

Hi Simone! I'm not sure when the updated Swift SDK is going to be available. The plumbing for a tar-based implementation will be the same regardless of whether we have a homegrown tar or libarchive, so I'm fine with merging everything we need for `cp` support using tar archives.

The main concern I have is that if we implement tar ourselves, it needs to be secure.

Sources/Containerization/Vminitd.swift
```diff
@@ -528,6 +530,159 @@ extension Vminitd {
...
+            } else {
+                let fd = try FileDescriptor.open(FilePath(item.path), .readOnly)
+                defer { try? fd.close() }
+                try writer.writeFile(path: entryPath, from: fd, mode: mode ?? 0o644)
+            }
+        }
+    }
+
+    /// Extract a tar archive to a destination directory.
+    private static func extractTar(reader: TarReader, to destURL: URL) throws {
```
jglogan, line 654: @simone-panico Have a look at https://github.com/apple/containerization/blob/main/Sources/ContainerizationArchive/ArchiveReader.swift#L304. I had to do similar fixes there back in January 😄.

You're skipping entries containing `..`, but the implementation can currently write through symlinks there's a TOCTOU race between creating the parent directory for a regular file extraction, and writing the regular file.

The tests for exercising safe extraction are at: https://github.com/simone-panico/containerization/blob/main/Tests/ContainerizationArchiveTests/ArchiveReaderTests.swift

jglogan:
Dev snapshots are out now:

```bash
swift sdk install \
    https://download.swift.org/swift-6.3-branch/static-sdk/swift-6.3-DEVELOPMENT-SNAPSHOT-2026-02-27-a/swift-6.3-DEVELOPMENT-SNAPSHOT-2026-02-27-a_static-linux-0.1.0.artifactbundle.tar.gz \
    --checksum 7dc6f58d1300438ebcf2e998de7c69d8da0ec42102460bd8e7f91c668e8b30c6
```

simone-panico:
> Dev snapshots are out now:
> 
> ```shell
> swift sdk install \
>     https://download.swift.org/swift-6.3-branch/static-sdk/swift-6.3-DEVELOPMENT-SNAPSHOT-2026-02-27-a/swift-6.3-DEVELOPMENT-SNAPSHOT-2026-02-27-a_static-linux-0.1.0.artifactbundle.tar.gz \
>     --checksum 7dc6f58d1300438ebcf2e998de7c69d8da0ec42102460bd8e7f91c668e8b30c6
> ```

Nice, I will try to switch out the `ContainerizatonTar` with `Libarchive`

dcantah:
@simone-panico Do you mind if we try and handle the Containerization side and you can plumb this through in `container`? There's quite a few pitfalls I keep thinking of in my head for this that are not simple to solve. 

1. There's not a great holding spot to write the archives to today as the VMs rootfs is read only. We can write to /tmp but this is a tmpfs, and if the archive was decently large we can run the risk of OOM. 
2. libarchive will be synchronous blocking IO, and we need to find a decent approach to run this without stalling the guest (as a user can choose a 1 core guest if they really want). Likely farm the work off to a thread and bridge with a continuation, but the devil is in the details.

There's a lot of solutions to these, but they all have drawbacks to me..

simone-panico:
> @simone-panico Do you mind if we try and handle the Containerization side and you can plumb this through in `container`? There's quite a few pitfalls I keep thinking of in my head for this that are not simple to solve.
> 
> 1. There's not a great holding spot to write the archives to today as the VMs rootfs is read only. We can write to /tmp but this is a tmpfs, and if the archive was decently large we can run the risk of OOM.
> 2. libarchive will be synchronous blocking IO, and we need to find a decent approach to run this without stalling the guest (as a user can choose a 1 core guest if they really want). Likely farm the work off to a thread and bridge with a continuation, but the devil is in the details.
> 
> There's a lot of solutions to these, but they all have drawbacks to me..

Yeah sure, go for it. The main goal in the first place for me was to finally merge the `container` side 😄

PR #568, opened 2026-03-03
LinuxContainer/LinuxPod: Fix UDS proxy if target is symlink

### LINKED ISSUE #256: [Bug]: Can't mount a host socket when the mount target goes through a symlink.

dcantah:
Closes #256

If the target of the uds proxy goes through a symlink (/var/run) we had the same bug we had with mounts previously where we'd follow this and if the resolved path existed in the root of the VM we'd end up mounting there. Fix this by plopping the sockets in a holding spot and then bind mounting them in (which has the correct resolution logic).

### DIFF
--- a/Sources/Containerization/LinuxPod.swift
+++ b/Sources/Containerization/LinuxPod.swift
@@ -254,6 +254,10 @@ public final class LinuxPod: Sendable {
     private static func guestRootfsPath(_ containerID: String) -> String {
         "/run/container/\(containerID)/rootfs"
     }
+
+    private static func guestSocketStagingPath(_ containerID: String, socketID: String) -> String {
+        "/run/container/\(containerID)/sockets/\(socketID).sock"
+    }
 }
 
 extension LinuxPod {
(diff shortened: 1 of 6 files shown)


Sources/Containerization/LinuxPod.swift
```diff
@@ -254,6 +254,10 @@ public final class LinuxPod: Sendable {
     private static func guestRootfsPath(_ containerID: String) -> String {
         "/run/container/\(containerID)/rootfs"
     }
+
+    private static func guestSocketStagingPath(_ containerID: String, socketID: String) -> String {
```
jglogan, line 258: Not necessary today, but at some point DRY this out by extracting this onto a UnixSocketConfiguration extension? Maybe there's other duplicate code between pod and container that can benefit from this treatment.

dcantah: It's likely we could get some benefit for how single files are setup as they're similar

PR #570, opened 2026-03-04
vminitd: Bump Swift version to 6.3-snapshot

### LINKED ISSUE #221: [Request]: Explore different allocators for vminitd

Musl's allocator is fairly poor in multithreaded scenarios. We should explore mimalloc/jemalloc/etc. if it's not too much hassle. 

### Code of Conduct

dcantah:
Closes #221

This contains (in the static linux sdk but the swift version and the sdk must match) libarchive support and mimalloc! This also (unrelated to the sdk bump it seems) raises the amount of retries we do when dialing vminitd. It seems CI on rare occasions recently has timed out on some tests. We could also explore just lowering the concurrency on the integration suite.


dcantah:
Hang on, investigating CI failures

dcantah:
Looks like the ci failure was unrelated to this change. My hunch is the integration tests are causing high cpu load and the first agent conn is timing out. I added a small diff here to try 50 more times

vminitd/Package.swift
```diff
@@ -53,6 +53,7 @@ let package = Package(
                 .product(name: "Containerization", package: "containerization"),
                 .product(name: "ContainerizationNetlink", package: "containerization"),
                 .product(name: "ContainerizationIO", package: "containerization"),
+                .product(name: "ContainerizationArchive", package: "containerization"),
```
dkovba: This change is out of scope of this PR.

Sources/Containerization/VZVirtualMachine+Helpers.swift
```diff
@@ -134,7 +134,7 @@ extension VZVirtualMachine {
                 continue
             }
         }
-        throw ContainerizationError(.invalidArgument, message: "no connection to agent socket")
+        throw ContainerizationError(.invalidArgument, message: "failed to get a connection to agent socket")
```
dkovba: This should be `.timeout` instead of `.invalidArgument`.

PR #571, opened 2026-03-05
LinuxContainer: Support copying directories

dcantah:
Now that the static linux SDK we use has libarchive linked against it, we can finally live our dream (/s) of supporting copying directories in/out a little easier. This was.. kind of annoying. I think the very simple route of writing archive to guest/host -> streaming over grpc is simplest, but on the guest end we have a couple problems:

1. The VMs rootfs is read only today always, which is a good thing to me and I don't want to change if we don't have to.
2. Writing the archive to the containers rootfs temporarily could work, but it's a bit weird, and the user can make the containers rootfs readonly which would screw that plan.
3. We could write it to /run or /tmp, but they're tmpfs and dealing with the headache of the user possibly tarring an enormous dir is one I don't want to care about.

So, that leaves us with the (truthfully better to me) approach of trying to write the tar data directly to the host and skipping grpc which kinda forces us to have a temp spot. Because of that, I made it such that we pass a port number from host<->guest, and transfer the actual data (either the single file or tarred dir contents) over the vsock port instead. The stream is kinda clunky, but it just serves as a means to exchange metadata and a "we're done" signifier.


Sources/Containerization/Vminitd.swift
```diff
@@ -439,92 +439,52 @@ extension Vminitd {
...
+        public let totalSize: UInt64
+    }
+
+    /// Unified copy control plane. Sends a CopyRequest over gRPC and processes
+    /// the response stream. Data transfer happens over a separate vsock connection
+    /// managed by the caller.
+    ///
+    /// For COPY_OUT, the `onMetadata` callback is invoked when the guest sends
+    /// metadata (is_archive, total_size) before data transfer begins.
+    /// For COPY_IN, `onMetadata` is not called.
```
JaewonHur, line 456: Isn't it called anyway? Couldn't find different path for `COPY_IN`.

PR #578, opened 2026-03-12
Upgrade vminitd API to use grpc-swift-2.

jglogan:
- No real notable API changes. The largest change code-wise is in Vminitd. The connection loop is set up immediately in the new API by calling runConnections() on the client, and this blocks until either the task it's running on is cancelled or beginGracefulShutdown is called. Because of this, either we'd have to modify our API to have VirtualMachineAgent implementations call some run() method, or we just put runConnections in an internal task on Vminitd. This change uses the latter approach as it's simpler.
- Updated dependencies to pick up latest protobuf and grpc-swift-2 fixes.
- ClientBootstrap.withConnectedSocket(fd).wait() registers the vsock fd with epoll/kqueue immediately. vminitd sends its HTTP/2 SETTINGS frame right away upon accepting the connection, so those bytes can arrive on the fd — and be read by NIO — before the gRPC pipeline is installed. With a bare pipeline, NIO has nowhere to send the bytes and discards them.
- Buffer until ClientConnectionHandler is in the pipeline so it can process the SETTINGS frame and fire .ready, then schedule delivery of the buffered data with assumeIsolatedUnsafeUnchecked().execute.
- To help produce more accurate merged logs, defines a `StderrLogHandler` in the vminitd Application.swift that writes timestamps to the boot log with millisecond precision and format identical to that of the log files created using the `--log-root` arg on `container`.
- Updates hawkeye to latest version (v6.5.1).

### DIFF
--- a/Sources/Containerization/SandboxContext/SandboxContext.grpc.swift
+++ b/Sources/Containerization/SandboxContext/SandboxContext.grpc.swift
@@ -45,7 +45,8 @@ public enum Com_Apple_Containerization_Sandbox_V3_SandboxContext: Sendable {
             /// Descriptor for "Mount".
             public static let descriptor = GRPCCore.MethodDescriptor(
                 service: GRPCCore.ServiceDescriptor(fullyQualifiedService: "com.apple.containerization.sandbox.v3.SandboxContext"),
-                method: "Mount"
+                method: "Mount",
+                type: .unary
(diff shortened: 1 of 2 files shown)


dcantah:
Based off of https://github.com/apple/containerization/pull/379 which we can close

PR #579, opened 2026-03-13
Enforce progress event value types

dkovba:
Enforces progress event value types.

### DIFF
--- a/Sources/ContainerizationExtras/ProgressEvent.swift
+++ b/Sources/ContainerizationExtras/ProgressEvent.swift
@@ -14,39 +14,35 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
-/// The kind of a progress update event.
-public enum ProgressEventKind: Sendable {
-    case addItems(Int)
-    case addTotalItems(Int)
-    case addSize(Int64)
-    case addTotalSize(Int64)
-}
-
 /// A progress update event.
-public struct ProgressEvent: Sendable {
-    /// The event name. The possible values:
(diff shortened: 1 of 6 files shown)


Sources/ContainerizationExtras/ProgressEvent.swift
```diff
@@ -17,10 +17,10 @@
...
 public struct ProgressEvent: Sendable {
     /// The event name. The possible values:
-    ///  - `add-items`: Increment the number of processed items by `value`.
-    ///  - `add-total-items`: Increment the total number of items to process by `value`.
-    ///  - `add-size`: Increment the size of processed items by `value`.
-    ///  - `add-total-size`: Increment the total size of items to process by `value`.
+    ///  - `add-items`: Increment the number of processed items by `value` (`Int`).
+    ///  - `add-total-items`: Increment the total number of items to process by `value` (`Int`).
+    ///  - `add-size`: Increment the size of processed items by `value` (`Int64`).
+    ///  - `add-total-size`: Increment the total size of items to process by `value` (`Int64`).
```
dcantah: If we can only provide known strings, and the types of the values are fixed as well, why don't we just make them those types?

dcantah: (e.g. some enum for the event, and an Int64 across the board for the values. Was this future proofing?)

dkovba: We used string-value pairs instead of an enum on purpose. However, it's still possible to use an enum for enforcing value types.

PR #581, opened 2026-03-13
LinuxContainer/LinuxPod: Add support for NoNewPrivileges

dcantah:
Basically just a direct mapping to the prctl/OCI spec field.


Sources/Containerization/LinuxProcessConfiguration.swift
```diff
@@ -378,6 +378,10 @@ public struct LinuxProcessConfiguration: Sendable {
     /// behavior and are planning to use a terminal for stdin/out/err on the client side,
     /// this should likely be set to true.
     public var terminal: Bool = false
+    /// Whether to set the no_new_privileges bit on the container process. When true, the
+    /// process and its children cannot gain additional privileges via setuid/setgid binaries
+    /// or file capabilities.
+    public var noNewPrivileges: Bool = false
```
dkovba: Consider placing before `capabilities` to match the order in the OCI specs.

PR #584, opened 2026-03-14
Network: Move to standalone type

dcantah:
The network protocol and VMNetNetwork implementation currently are housed on ContainerManager even though they are generally useful types even outside of this easy to use helper type. This change moves them to not be nested types anymore, as well as exposes a new param on VMNetNetworks constructor so we can pass the type of network.

### DIFF
--- a/Sources/Containerization/Network.swift
+++ b/Sources/Containerization/Network.swift
@@ -1,0 +1,21 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the Containerization project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// A network that can allocate and release interfaces for use with containers.
+public protocol Network: Sendable {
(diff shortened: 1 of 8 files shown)


Sources/Containerization/Network.swift
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// A network that can allocate and release interfaces for use with containers.
+public protocol Network: Sendable {
+    mutating func create(_ id: String) throws -> Interface?
```
jglogan: Create what? A network or an interface? IMO this should be `createInterface` and `releaseInterface` if we're creating public APIs for this.

Sources/Containerization/VmnetNetwork.swift
```diff
@@ -0,0 +1,209 @@
...
+        let cidr = try Self.getSubnet(ref)
+
+        self.allocator = try .init(cidr: cidr)
+        self.subnet = cidr
+        self.reference = ref
+    }
+
+    /// Returns a new interface for use with a container.
+    /// - Parameter id: The container ID.
+    public mutating func create(_ id: String) throws -> Containerization.Interface? {
```
jglogan: It seems odd that we define the protocol and this conforms to it, and then we've got a couple extra `create` functions that don't.

I guess the question here is: what's the point of having a protocol.

dcantah: It's so ContainerManager can take in anything, and poop out Interface's. But for the more "in control" types like LinuxContainer you can call any of these. ContainerManager is meant to fulfill the very simplest use cases.

Sources/Containerization/Network.swift
dcantah: Yea good call, this is too generic. Will fix in a little

PR #589, opened 2026-03-17
Write archive contents in chunks

adityaramani:
Fix a bug in ArchiveWriter where large files would only partly be written into an archive. 
This is done so by writing the data in chunks.

### DIFF
--- a/Sources/ContainerizationArchive/ArchiveWriter.swift
+++ b/Sources/ContainerizationArchive/ArchiveWriter.swift
@@ -282,7 +282,7 @@ extension ArchiveWriter {
                 }
                 defer { buf.deallocate() }
                 let fd = Foundation.open(fullPath.string, O_RDONLY)
-                guard fd > 0 else {
+                guard fd >= 0 else {
                     let err = POSIXErrorCode(rawValue: errno) ?? .EINVAL
                     throw ArchiveError.failedToCreateArchive("cannot open file \(fullPath.string) for reading: \(err)")
                 }


Sources/ContainerizationArchive/ArchiveWriter.swift
```diff
@@ -267,8 +274,20 @@ extension ArchiveWriter {
...
-                let data = try Data(contentsOf: URL(fileURLWithPath: fullPath.string), options: .uncached)
-                try self.writeEntry(entry: entry, data: data)
+                guard let fileHandle = FileHandle(forReadingAtPath: fullPath.string) else {
+                    throw POSIXError(.ENOENT)
+                }
+                defer { fileHandle.closeFile() }
+                try self.writeHeader(entry: entry)
+                let chunkSize = 4 * 1024 * 1024
+                while true {
+                    let chunk = fileHandle.readData(ofLength: chunkSize)
```
dcantah: Can we just have a preallocated scratch buffer in our writer of size `chunkSize` that we fill, and use read(2) directly. I have a vendetta against fileHandle readData LOL

adityaramani: updated

Sources/ContainerizationArchive/ArchiveWriter.swift
```diff
@@ -267,8 +276,28 @@ extension ArchiveWriter {
...
             if type == .regular {
-                let data = try Data(contentsOf: URL(fileURLWithPath: fullPath.string), options: .uncached)
-                try self.writeEntry(entry: entry, data: data)
+                let buf = UnsafeMutableRawBufferPointer.allocate(byteCount: Self.chunkSize, alignment: 1)
+                guard let baseAddress = buf.baseAddress else {
+                    throw ArchiveError.failedToCreateArchive("cannot create temporary buffer of size \(Self.chunkSize)")
+                }
+                defer { buf.deallocate() }
+                let fd = Foundation.open(fullPath.string, O_RDONLY)
+                guard fd > 0 else {
```
manojmahapatra: Question: should not this be `fd >= 0`?

Sources/ContainerizationArchive/ArchiveReader.swift
```diff
@@ -375,7 +375,7 @@ extension ArchiveReader {
     }
 
     private static func copyDataReaderToFd(dataReader: ArchiveEntryReader, fileFd: Int32, memberPath: FilePath) throws {
-        var buffer = [UInt8](repeating: 0, count: ArchiveReader.blockSize)
+        var buffer = [UInt8](repeating: 0, count: ArchiveReader.chunkSize)
```
manojmahapatra, line 378: Question: Does `copyDataReaderToFd` need to handle partial writes by looping until the full chunk is written?

Sources/ContainerizationArchive/ArchiveWriter.swift
adityaramani: yep you are right. Fixed

Sources/ContainerizationArchive/ArchiveReader.swift
adityaramani: It does that already right?

PR #590, opened 2026-03-18
Fix createParents in vminitd copy handler

simone-panico:
Respect `createParents` flag when creating the destination directory for archive extraction.

### DIFF
--- a/vminitd/Sources/vminitd/Server+GRPC.swift
+++ b/vminitd/Sources/vminitd/Server+GRPC.swift
@@ -462,7 +462,7 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
                 return []
             }
             let destURL = URL(fileURLWithPath: path)
-            try FileManager.default.createDirectory(at: destURL, withIntermediateDirectories: true)
+            try FileManager.default.createDirectory(at: destURL, withIntermediateDirectories: request.createParents)
 
             let fileHandle = FileHandle(fileDescriptor: sockFd, closeOnDealloc: false)
             let reader = try ArchiveReader(format: .pax, filter: .gzip, fileHandle: fileHandle)


vminitd/Sources/vminitd/Server+GRPC.swift
```diff
@@ -462,7 +462,7 @@ extension Initd: Com_Apple_Containerization_Sandbox_V3_SandboxContextAsyncProvid
                 return []
             }
             let destURL = URL(fileURLWithPath: path)
-            try FileManager.default.createDirectory(at: destURL, withIntermediateDirectories: true)
+            try FileManager.default.createDirectory(at: destURL, withIntermediateDirectories: request.createParents)
```
JaewonHur, line 465: I feel like this should remain `withIntermediateDirectories: true`. We already make parent directories in L414. Directory here is to unpack the tar archive.

simone-panico: I can't do the retry-based approach in [container#1190](https://github.com/apple/container/pull/1190) without this because the CLI needs the first `copyIn(src, dst/basename, createParents=false)` to fail if dst doesn't exist. 
If `withIntermediateDirectories` is always true, the createDirectory creates dst as a side effect, the copy lands at the wrong path and the error-based retry never triggers

JaewonHur: Yes. We are stuck :(
Sorry for taking up your time. We'll think it over a bit more and let you know once we've decided on a direction.

simone-panico: No worries, take your time. Just let me know once you've settled on a direction :)

crosbymichael: What bug are you fixing with this? This is an interesting change and I can see why we are stuck. I would expect the request bool for creating parents to only apply to the initial parent directory, not sub-directories.

simone-panico: This was initially for this PR [container#1190](https://github.com/apple/container/pull/1190).
But we found another solution and I think we can close this PR, right @JaewonHur ?

PR #593, opened 2026-03-23
Add Seccomp support

### LINKED ISSUE #551: [Request]: expose seccomp in the API

Hi folks!

I'd like a way to specify a `noNewPrivileges: Bool` and a `seccomp: LinuxSeccomp?` property, so that callers of this library can pass those settings through to the OCI runtime spec that gets executed inside the VM.

Both fields already exist in `ContainerizationOCI` (`Process.noNewPrivileges` and
`Linux.seccomp`) — the missing piece is that the higher-level Swift API doesn't surface them, so library consumers have no way to set them.

My plan would be to have a way to set these attributes when generating the OCI image, and then on top of that make changes to the `apple/container` repo so that it makes good use of them, and expose it.

## Why this matters — the Dangerzone use case

[Dangerzone](https://github.com/freedomofpress/dangerzone) is a security tool from [Freedom of the Press Foundation](https://freedom.press) that converts potentially malicious documents (PDFs, Office files, images) into safe PDFs by processing them inside an isolated container. The threat model is explicit: the document being converted may be hostile, crafted specifically to escape the sandbox. Keeping the container process as locked-down as possible is therefore not optional — it is the whole point of the tool.

Dangerzone runs on Linux (Docker/Podman), and we've started experimenting `apple/container` support for macOS, which relies on this library. On every other container runtime it uses the following security flags unconditionally:

```python
security_args = ["--security-opt", "no-new-privileges"]
security_args += ["--security-opt", f"seccomp={seccomp_json_path}"]
security_args += ["--cap-drop", "all"]
security_args += ["--cap-add", "SYS_CHROOT"]
```

Without `noNewPrivileges` and a custom seccomp profile, a container process that is compromised by a malicious document could potentially escalate privileges or make syscalls that are not needed for legitimate rendering work, widening the attack surface. These two controls are the last line of defense after capability dropping.

Hope it makes sense! I believe we could add this behavior in a way that doesn't make changes to the default behavior, but I have limited understanding of the current codebase, and so let me know if I'm missing something!

Thanks!

### Code of Conduct

dcantah:
Closes #551

This change is rather large, but I think this is one of those changes that is much simpler to see all laid out. It:

- Adds a new ContainerizationSeccomp target/product that is a cBPF compiler specifically for seccomp. Its main use is to take in an OCI seccomp description and spit out a filter we can apply.
- Adds a new friendly SeccompProfile API to Containerization to specify what filters you'd like applied. This will (as is the case for basically everything else) get translated to OCI behind the scenes.
- Adds a small bit of logic in vmexec to apply the filters.

And unit and integration tests for everything. Unit testing is interesting. I've added a small simulator so we actually have some semblance of testing outside of just integration tests and seeing if the syscall is blocked/returns whatever.

Right now the codegen is somewhat simple. I'm sure it can be better, but it's a start.


almet:
Hi, thanks for working on this! Let me know if there is anything I can do to help making this happen :-)

dcantah:
@almet I'll try and get some folks to take a gander in the coming weeks. It's a rather large surface

almet:
Hey, here is a gentle ping, in case it can help move things forward.

I believe we could start to use containers for the [dangerzone](https://dangerzone.rocks) project soon, in order to replace podman on macOS, but we currently rely on using a custom seccomp policy for our containers.

Do you know if there is anything we could do to help? it would be great!

sturdy4days:
I reviewed the current PR head while tracking #551 for a downstream security consumer. Before this is merged, could the compiler fail closed for unresolved syscall names and add focused coverage for that behavior?

At the current head, an unresolved name falls through with `continue`, and the existing test documents that an unknown name emits no rule. That is unsafe for a default-allow denylist: a typo, profile drift, or a newly named syscall can silently remove the intended deny rule and leave the workload unconfined for that syscall.

A narrow compatible direction would be:

- return the existing `unknownSyscall` error for unresolved names;
- retain `.number(UInt32)` for deliberately raw/new syscall numbers;
- add unknown-only and mixed known/unknown tests under default-allow, plus a startup/integration assertion that an unresolved requested rule prevents profile installation/workload startup rather than proceeding unfiltered.

The PR also appears aarch64-only and is currently dirty against main. I would keep x86_64 expansion as a separately scoped change unless you prefer to broaden this PR. Happy to contribute a non-overlapping follow-up after the intended ownership/rebase path is clear.

PR #596, opened 2026-03-23
Remove the `macos-sdk` build target

dkovba:
Removes the `macos-sdk` build target.


dcantah:
Dumb question, do we even need a macOS sdk for vminitd builds?

dkovba:
> Dumb question, do we even need a macOS sdk for vminitd builds?

No, we don't. Removed the `macos-sdk` target from this `Makefile` completely.

PR #601, opened 2026-03-27
LinuxContainer/LinuxPod: Sort mounts by depth

dcantah:
To avoid a user accidentally shadowing a mount, we can sort mounts by depth (but stable sort so mounts of the same depth stay in the order provided).

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -22,6 +22,7 @@ import ContainerizationOCI
 import Foundation
 import Logging
 import Synchronization
+import SystemPackage
 
 import struct ContainerizationOS.Terminal
 
@@ -671,7 +672,7 @@ extension LinuxContainer {
                         ))
                 }
 
-                spec.mounts = mounts
+                spec.mounts = cleanAndSortMounts(mounts)
 
                 let stdio = IOUtil.setup(
                     portAllocator: self.hostVsockPorts,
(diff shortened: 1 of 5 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -1292,6 +1292,16 @@ extension AttachedFilesystem {
     }
 }
 
+/// Sort mounts by the depth of their destination path. This ensures that
+/// higher level mounts don't shadow other mounts. For example, if a user
+/// specifies mounts for `/tmp/foo/bar` and `/tmp`, sorting by depth ensures
+/// `/tmp` is mounted first without shadowing `/tmp/foo/bar`.
+func sortMountsByDestinationDepth(_ mounts: [ContainerizationOCI.Mount]) -> [ContainerizationOCI.Mount] {
```
manojmahapatra, line 1310: Question: Could odd paths like `/tmp/./foo` or `/tmp/../foo` reach here?

dcantah: Yea a user can pass anything. We should normalize

jglogan: Perhaps `ContainerOCI.Mount` should perform lexical normalization on paths? Is there any reason that a mount path should store an unnormalized path?

Or, we do lexical normalization here only for the purpose of depth computation.

dcantah: I'd prefer to keep the OCI/config types dumb and just be configuration. We can "fixup" the mounts before we ship them to the guest (normalize, sort etc.).

jglogan: I could imagine symlinks being able to fool the depth check as well (though in cases like that perhaps you deserve what you get):

With `/foo/bar/baz/qux` -> `/opt`, `/foo/bar/baz/qux/mnt` would sort after `/opt/mnt/data`.

I know there are a ton of crazy cases but is there any way where we can "just mount in the order the user told us" and if things get shadowed, then they should use the correct order themselves?

dcantah: Normalized now

dcantah: There will always be a certain aspect of "the user can pass us complete garbage" that we can't program around, we have to just bubble up the error from the guest. But the normalizing on /whatever/../thing is a good idea.

manojmahapatra: yeah, could not agree more.

PR #604, opened 2026-03-30
Add the vminitd version to the logs

dkovba:
Adds the vminitd version to the logs.

Switching to a local version of vminitd is non-trivial:

```
-                .define("CZ_VERSION", to: "\"\(scVersion)\""),
+                .define("CZ_VERSION", to: "\"latest\""),
```

Having an additional confirmation of the used version in the logs is vital.

### DIFF
--- a/vminitd/Sources/vminitd/AgentCommand.swift
+++ b/vminitd/Sources/vminitd/AgentCommand.swift
@@ -67,7 +67,7 @@ struct AgentCommand: AsyncParsableCommand {
 
         signal(SIGPIPE, SIG_IGN)
 
-        log.info("vminitd booting")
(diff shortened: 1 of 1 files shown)


dcantah:
Wait, I just realized this is just referencing the Swift arg parser version which we haven't changed at all.. It's been 0.1 since launch. Should we use git sha for a better diagnostic..

dkovba:
> Wait, I just realized this is just referencing the Swift arg parser version which we haven't changed at all.. It's been 0.1 since launch. Should we use git sha for a better diagnostic..

A git SHA wouldn't reflect local changes. What do you think about [logging the binary's modification date and time](https://github.com/apple/containerization/pull/608) instead?

PR #614, opened 2026-03-31
Add stat RPC

JaewonHur:
Add `Stat` RPC to vminitd.

`Stat` RPC is used to stat file system objects in the VM. It accepts the path to the object (resolved against VM root) and returns the stat result.


Sources/Containerization/Vminitd.swift
```diff
@@ -448,6 +448,21 @@ extension Vminitd {
         public let totalSize: UInt64
     }
 
+    public func stat(
+        path: URL
+    ) async throws -> Com_Apple_Containerization_Sandbox_V3_Stat {
```
dcantah: Can we return a stat_t? And not the protobuf type directly?

Sources/Containerization/Vminitd.swift
```diff
@@ -448,6 +448,36 @@ extension Vminitd {
...
+        let s = response.stat
+        var result = Foundation.stat()
+        result.st_dev = dev_t(s.dev)
+        result.st_mode = mode_t(s.mode)
+        result.st_nlink = nlink_t(s.nlink)
+        result.st_ino = ino_t(s.ino)
+        result.st_uid = s.uid
+        result.st_gid = s.gid
+        result.st_rdev = dev_t(s.rdev)
+        // result.st_atimespec = timespec(tv_sec: Int(s.atime.seconds), tv_nsec: Int(s.atime.nanos))
```
JaewonHur: @dcantah updated to return `Foundation.stat`.

But, it doesn't compile when I set `timespec` fields there. It feels like because they are defined differently depending on the underlying OS.

```
#define __DARWIN_STRUCT_STAT64_TIMES \
	struct timespec st_atimespec;           /* time of last access */ \
	struct timespec st_mtimespec;           /* time of last data modification */ \
	struct timespec st_ctimespec;           /* time of last status change */ \
	struct timespec st_birthtimespec;       /* time of file creation(birth) */

#else /* (_POSIX_C_SOURCE && !_DARWIN_C_SOURCE) */

#define __DARWIN_STRUCT_STAT64_TIMES \
	time_t		st_atime;               /* [XSI] Time of last access */ \
	long		st_atimensec;           /* nsec of last access */ \
	time_t		st_mtime;               /* [XSI] Last data modification time */ \
	long		st_mtimensec;           /* last data modification nsec */ \
	time_t		st_ctime;               /* [XSI] Time of last status change */ \
	long		st_ctimensec;           /* nsec of last status change */ \
	time_t		st_birthtime;           /*  File creation time(birth)  */ \
	long		st_birthtimensec;       /* nsec of File creation time */

#endif /* (!_POSIX_C_SOURCE || _DARWIN_C_SOURCE) */

/*
 * This structure is used as the second parameter to the fstat64(),
 * lstat64(), and stat64() functions, and for struct stat when
 * __DARWIN_64_BIT_INO_T is set. __DARWIN_STRUCT_STAT64 is defined
 * above, depending on whether we use struct timespec or the direct
 * components.
 *
 * This is simillar to stat except for 64bit inode number
 * number instead of 32bit ino_t and the addition of create(birth) time.
 */
#define __DARWIN_STRUCT_STAT64 { \
	dev_t		st_dev;                 /* [XSI] ID of device containing file */ \
	mode_t		st_mode;                /* [XSI] Mode of file (see below) */ \
	nlink_t		st_nlink;               /* [XSI] Number of hard links */ \
	__darwin_ino64_t st_ino;                /* [XSI] File serial number */ \
	uid_t		st_uid;                 /* [XSI] User ID of the file */ \
	gid_t		st_gid;                 /* [XSI] Group ID of the file */ \
	dev_t		st_rdev;                /* [XSI] Device ID */ \
	__DARWIN_STRUCT_STAT64_TIMES \
	off_t		st_size;                /* [XSI] file size, in bytes */ \
	blkcnt_t	st_blocks;              /* [XSI] blocks allocated for file */ \
	blksize_t	st_blksize;             /* [XSI] optimal blocksize for I/O */ \
	__uint32_t	st_flags;               /* user defined flags for file */ \
	__uint32_t	st_gen;                 /* file generation number */ \
	__int32_t	st_lspare;              /* RESERVED: DO NOT USE! */ \
	__int64_t	st_qspare[2];           /* RESERVED: DO NOT USE! */ \
}

/*
 * [XSI] This structure is used as the second parameter to the fstat(),
 * lstat(), and stat() functions.
 */
#if __DARWIN_64_BIT_INO_T

struct stat __DARWIN_STRUCT_STAT64;
```

Do you have any workaround?

dcantah: Can we define our own stat type that has the relevant fields we care about and just populate them and return that type?

dcantah (approved):

PR #619, opened 2026-03-31
Epoll: Rework epoll type

dcantah:
I've spent a lot of time in the past trying to fix all of the various issues with the epoll wrapper we had, before realizing most of them would be completely gone if we just reworked the type itself :). The callback nature, all of the handler/state tracking internally all (to me) has no purpose being in the type itself. All of this logic can live outside and the wrapper should just be a typesafe abstraction around it that you can build on. This is what this change aims to do. There should be zero behavioral difference here and this gets rid of quite a few bugs (and makes things compile with glibc)

### DIFF
--- a/Sources/ContainerizationOS/Linux/Epoll.swift
+++ b/Sources/ContainerizationOS/Linux/Epoll.swift
@@ -15,16 +15,17 @@
 //===----------------------------------------------------------------------===//
 
 #if os(Linux)
+import Foundation
 
 #if canImport(Musl)
 import Musl
+private let _write = Musl.write
 #elseif canImport(Glibc)
 import Glibc
+private let _write = Glibc.write
 #endif
 
 import CShim
-import Foundation
-import Synchronization
 
 // On glibc, epoll constants are EPOLL_EVENTS enum values. On musl they're
 // plain UInt32. These helpers normalize them to UInt32/Int32.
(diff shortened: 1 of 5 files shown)


Sources/ContainerizationOS/Linux/Epoll.swift
```diff
@@ -14,168 +14,178 @@
...
-        if result != 0 {
-            if !acceptableDeletionErrno() {
-                throw POSIXError.fromErrno()
-            }
-        }
-        self.handlers.del(fd)
+    /// Signal the epoll loop to stop waiting.
+    public func shutdown() {
+        var byte: UInt8 = 0
+        _ = Musl.write(shutdownWriteFD, &byte, 1)
```
yibozhuang: Should this be not specific to Musl and just be `write()`

Sources/ContainerizationOS/Linux/Epoll.swift
```diff
@@ -14,168 +14,178 @@
...
         while true {
             let n = epoll_wait(self.epollFD, &events, Int32(events.count), timeout)
-            guard n >= 0 else {
+            if n < 0 {
                 if errno == EINTR || errno == EAGAIN {
-                    continue  // go back to epoll_wait
+                    continue
                 }
-                throw POSIXError.fromErrno()
+                return []
```
yibozhuang: Do we want to just return empty if `epoll_wait()` happens to error with `EBADF` or `EFAULT`? What's the reason we don't want to throw error here?

yibozhuang (approved):

PR #623, opened 2026-03-31
ArchiveReader: Cleanup tempdirs/files

dcantah:
There was a couple spots we were leaving around tempdirs

### DIFF
--- a/Sources/ContainerizationArchive/ArchiveReader.swift
+++ b/Sources/ContainerizationArchive/ArchiveReader.swift
@@ -77,8 +77,15 @@ public final class ArchiveReader {
             actualFilter = filter
         }
 
-        let fileHandle = try FileHandle(forReadingFrom: fileToRead)
-        try self.init(format: format, filter: actualFilter, fileHandle: fileHandle)
+        do {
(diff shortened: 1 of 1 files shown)


Sources/ContainerizationArchive/ArchiveReader.swift
```diff
@@ -233,9 +245,21 @@ extension ArchiveReader: Sequence {
...
+        }
+        let url = tempDir.appendingPathComponent(name)
+        do {
+            try bundle.write(to: url, options: .atomic)
+            try self.init(format: .zip, filter: .none, file: url)
+        } catch {
+            try? FileManager.default.removeItem(at: tempDir)
+            throw error
+        }
+        // Register for cleanup in deinit (only needed when the zstd path didn't already set it)
```
jglogan, line 259: We added the tempDecompressedFile specifically because we needed to try to uncompress zstd before invoking libarchive (with either the uncompressed zstd file or the original file if not zstd).

Here it looks like we're writing ZIP-format data into a temp directory created by the call but subsequently owned by the user when init returns? If the data's in the temp directory, why would we need to set tempDecompressedFile for this case?

PR #628, opened 2026-04-01
Support a default route without a gateway

dkovba:
Adds the support for a default route without a gateway.

```
/ # ip route show default
default dev eth0 scope link 
```


Sources/Containerization/LinuxPod.swift
```diff
@@ -457,6 +457,9 @@ extension LinuxPod {
                                 try await agent.routeAddLink(name: name, dstIPv4Addr: ipv4Gateway, srcIPv4Addr: nil)
                             }
                             try await agent.routeAddDefault(name: name, ipv4Gateway: ipv4Gateway)
+                        } else {
```
dcantah, line 460: We'll need to edit LinuxContainer as well

dkovba: Done

PR #629, opened 2026-04-02
Remove not used code which had a buffer overread

dkovba:
Removes not used `FilePath.init?(Data)` which had a buffer overread. It used `String(cString:)` which reads until `\0`, but `Data` is not null-terminated.


jglogan:
@dkovba test errors?

```
/Users/runner/actions-runner/_work/containerization/containerization/Tests/ContainerizationEXT4Tests/TestFormatterUnpack.swift:128:46: error: no exact matches in call to initializer 
126 |         let specialFileInode = try ext4.getInode(number: 20)
127 |         let bytes = Data(Mirror(reflecting: specialFileInode.block).children.compactMap { $0.value as? UInt8 })
128 |         let specialFileTarget = try #require(FilePath(bytes), "Could not parse special file path")
    |                                              `- error: no exact matches in call to initializer 
129 |         #expect(specialFileTarget.description.hasPrefix("special_ÆÂ©"))
130 |     }
```

PR #634, opened 2026-04-03
Write file data in chunks in ContentWriter

adityaramani:
Update the `ContentWriter.create(from url: URL)` method to not load the whole file in memory when computing its hash and writing its contents

### DIFF
--- a/Sources/ContainerizationOCI/Content/ContentWriter.swift
+++ b/Sources/ContainerizationOCI/Content/ContentWriter.swift
@@ -58,16 +58,18 @@ public class ContentWriter {
     public func create(from url: URL) throws -> (size: Int64, digest: SHA256.Digest) {
         let sourceFD = Foundation.open(url.path, O_RDONLY)
         guard sourceFD >= 0 else {
-            let err = POSIXErrorCode(rawValue: errno) ?? .EINVAL
-            throw ContainerizationError(.internalError, message: "failed to open \(url.path) for reading: \(err)")
+            let errCode = POSIXErrorCode(rawValue: errno) ?? .EINVAL
+            let err = POSIXError(errCode)
+            throw ContainerizationError(.internalError, message: "failed to open \(url.path) for reading", cause: err)
         }
         defer { close(sourceFD) }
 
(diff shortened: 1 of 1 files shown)


Sources/ContainerizationOCI/Content/ContentWriter.swift
```diff
@@ -56,8 +56,42 @@ public class ContentWriter {
...
+        }
+        let dest = try FileHandle(forWritingTo: tempURL)
+        var hasher = SHA256()
+        var totalSize: Int64 = 0
+        let chunkSize = 1024 * 1024  // 1 MiB
+        do {
+            while let chunk = try source.read(upToCount: chunkSize), !chunk.isEmpty {
+                hasher.update(data: chunk)
+                try dest.write(contentsOf: chunk)
+                totalSize += Int64(chunk.count)
```
dcantah: Same ask as before, I'd love if we can avoid reading with FileHandle 🙏

adityaramani: Yep I keep forgetting. Updated!

Sources/ContainerizationOCI/Content/ContentWriter.swift
```diff
@@ -56,8 +56,70 @@ public class ContentWriter {
     ///   - url: The URL to read the data from.
     @discardableResult
     public func create(from url: URL) throws -> (size: Int64, digest: SHA256.Digest) {
-        let data = try Data(contentsOf: url)
-        return try self.write(data)
+        let sourceFD = Foundation.open(url.path, O_RDONLY)
+        guard sourceFD >= 0 else {
+            let err = POSIXErrorCode(rawValue: errno) ?? .EINVAL
+            throw ContainerizationError(.internalError, message: "failed to open \(url.path) for reading: \(err)")
```
dcantah: ContainerizationError has a `cause` field that we can put the posixerrors in

dcantah: So a client can actually programmatically handle whatever errno

PR #640, opened 2026-04-03
Remove the unreachable partial last block group handling in the EXT4 formatter

dkovba:
Removes the unreachable partial last block group handling in the EXT4 formatter.

### DIFF
--- a/Sources/ContainerizationEXT4/EXT4+Formatter.swift
+++ b/Sources/ContainerizationEXT4/EXT4+Formatter.swift
@@ -804,16 +804,7 @@ extension EXT4 {
                 inodeBitmap[Int(i) / 8] &= ~(1 << (i % 8))
             }
             for group in blockGroupSize.blockGroups..<totalGroups.lo {
-                var blocksInGroup = UInt32(self.blocksPerGroup)
-                if group == totalGroups.lo - 1 {
-                    if UInt64(self.size / UInt64(self.blockSize)) < self.blocksPerGroup {
-                        break
-                    }
-                    let lastGroupBlocks = UInt32((self.size / UInt64(self.blockSize)) % UInt64(self.blocksPerGroup))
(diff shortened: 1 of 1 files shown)


Sources/ContainerizationEXT4/EXT4+Formatter.swift
```diff
@@ -805,13 +805,13 @@ extension EXT4 {
             }
             for group in blockGroupSize.blockGroups..<totalGroups.lo {
```
wlan0, line 806: Please show the failing case this change fixes

The old comparison was definitely off by one for `for group in blockGroupSize.blockGroups..<totalGroups.lo`, because the last iterated value is `totalGroups.lo - 1`, not `totalGroups.lo`. 

But just above this loop, the formatter rounds `self.size` up to `totalGroups * blocksPerGroup * blockSize` whenever it is smaller. That makes me think `((self.size / blockSize) % blocksPerGroup)` is usually `0` by the time this branch runs, so `lastGroupBlocks` never changes `blocksInGroup`.

wlan0:
If the goal is to support a partial final group, the local bitmap change is not the whole story. 

The special case adjusts the last group bitmap and its `freeBlocksCountLow` via `blocksInGroup`, but later the superblock still derives `blocksCount` from `totalGroups * self.blocksPerGroup` and `totalFreeBlocks` from `blocksCount - totalBlocks`. 

In the used-group path, `freeBlocks` is computed and then not used. So if this partial-group path ever becomes reachable, the bitmap-level view and the global free-space counters could still differ.

Sources/ContainerizationEXT4/EXT4+Formatter.swift
dkovba: Removed the not executed code instead of the original partial fix. The complete fix can be done under https://github.com/apple/containerization/issues/647

PR #641, opened 2026-04-03
Add a guard against an empty range in the EXT4 formatter

dkovba:
Adds a guard against an empty range in the EXT4 formatter to prevent a possible crash:

> Swift/arm64e-apple-macos.swiftinterface:6314: Fatal error: Range requires lowerBound <= upperBound

### DIFF
--- a/Tests/ContainerizationEXT4Tests/TestEXT4Format.swift
+++ b/Tests/ContainerizationEXT4Tests/TestEXT4Format.swift
@@ -219,3 +219,14 @@ struct Ext4FormatTests: ~Copyable {
         #expect(regFile.sizeLow == 4)
     }
 }
+
+struct Ext4FormatEmptyRangeTests {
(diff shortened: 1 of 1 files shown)


wlan0:
I agree that this loop would need guarding if the lower bound could exceed the upper bound, but with the current math I can’t find a reachable state where that happens. 

Can you add a failing regression test or share the exact formatter input that makes 
`usedGroupDescriptorBlocks + 1 > self.groupDescriptorBlocks` ?

dkovba:
@wlan0 Added a test reproducing the crash. However, to reproduce the crash, I had to remove reserved blocks in `groupDescriptorBlocks`:
```
-            ((groupCount - 1) / groupsPerDescriptorBlock + 1) * 32                                           
+            (groupCount - 1) / groupsPerDescriptorBlock + 1 
```

PR #650, opened 2026-04-04
Add deep extent tree support (depth >= 2) in EXT4 formatter

gbfish:
## Summary
- **Reader**: Refactor `getExtents()` with recursive `readExtentNode()` that handles any depth
- **Formatter**: Extract `writeExtentSubtree()` that recursively builds extent trees of any depth
- Fix force-unwrap crash on `leafNode.leaves.last!.block` (use safe optional access)
- Replace hardcoded depth 0/1 switch with general depth calculation
- Consistent use of `.loadLittleEndian(as:)` for all on-disk struct reads
- Add test for multi-extent file readback

### DIFF
--- a/Sources/ContainerizationEXT4/EXT4+Formatter.swift
+++ b/Sources/ContainerizationEXT4/EXT4+Formatter.swift
@@ -1039,6 +1039,130 @@ extension EXT4 {
             }
         }
 
+        /// Recursively writes extent tree nodes to disk and returns the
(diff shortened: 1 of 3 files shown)


dcantah:
Thanks for the work. To me, would you be open to merging the two changes into one? It'd be simpler to look for logical bugs if the formatter and reader are all in one.

gbfish:
Done — combined both reader and formatter changes into this single PR. Thanks for the suggestion.

PR #652, opened 2026-04-05
ArchiveWriter: support explicit entries and rewrite internal symlinks

mazdak:
## Context and Motivation

While building out a Docker Compose-style plugin and validating it against our own real development stack, we ran into two classes of problems in the build-context path, and to fix them we had to make changes to the `container` app, but part of the changes were archive-authoring concerns and belong in `ContainerizationArchive`, not in higher-level callers. The result is a reusable API for explicit archive creation and more portable archives when symlinks point to other archived content.

(Related `container` PR: https://github.com/apple/container/pull/1391)

This PR adds a generic explicit-entry archive API to `ContainerizationArchive` and improves symlink handling so archives remain relocatable after extraction.

In addition to `archiveDirectory(_:)`, callers can now provide an ordered list of source entries to archive. This makes it possible to archive a precomputed file set without rewalking the source tree and centralizes symlink handling in the archive layer itself.

## Testing

  - add regression coverage for explicit-entry archive creation
  - add round-trip tests for:
    - internal absolute symlinks rewritten to archive-relative paths
    - external absolute symlinks preserved unchanged
    - canonicalized internal absolute symlinks through symlinked ancestors
  - verify existing archive-directory symlink behavior still passes

### Passed:
  - swift test --filter archiveEntries
  - swift test --filter archiveDirectoryRewritesInternalAbsoluteSymlink
  - swift test --filter archiveDirectorySymlinkInside
  - swift test --filter archiveDirectorySymlinkOutsideExcluded

### DIFF
--- a/Sources/ContainerizationArchive/ArchiveWriter.swift
+++ b/Sources/ContainerizationArchive/ArchiveWriter.swift
@@ -209,12 +209,8 @@ extension ArchiveWriter {
 extension ArchiveWriter {
     /// Archives an explicit, ordered list of host filesystem entries.
     public func archiveEntries(_ entries: [ArchiveSourceEntry]) throws {
-        let archivedPathsByHostPath = entries.reduce(into: [String: [String]]()) { result, entry in
-            result[entry.pathOnHost.path, default: []].append(entry.pathInArchive)
-        }
-
         for source in entries {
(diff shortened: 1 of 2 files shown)


jglogan:
@mazdak Sorry I've not looked closely enough at the changes yet, I've been swamped.

I'll get some feedback on this by Friday. I still need to look at the code more closely to see where this best fits.

PR #653, opened 2026-04-06
Add custom OOM killer for Linux containers

JaewonHur:
This PR implements a custom OOM killer that is spawned as a child process of `vmexec`.

While Linux kernel also OOM kills a process if it hits cgroup memory limit and the kernel cannot reclaim the memory, kernel often fails to kill the process and left the system hang due to the memory thrashing. Especially, the process is not OOM killed because the kernel still succeeds reclaiming the memory, not meeting the condition for OOM kill (but which takes way longer time, and leads to hang). 

Thus, this PR adds a user space OOM killer as a child process of `vmexec`, which monitors cgroup memory events, and kills the process when `max` event hits a specified limit. This approach can reliably kills the OOM process as monitoring memory events can be performed in small time window.

This PR needs following more works:
1. Plumb UI to inform the users that the container has been killed due to the OOM.
2. Refactor `errorPipe` to catch errors from (long running) OOM killer process (or any other ways to catch the errors).

### DIFF
--- a/vminitd/Sources/vmexec/RunCommand.swift
+++ b/vminitd/Sources/vmexec/RunCommand.swift
@@ -33,6 +33,9 @@ struct RunCommand: ParsableCommand {
     var bundlePath: String
 
     mutating func run() throws {
+        let syncPipe = FileDescriptor(rawValue: 3)
+        let ackPipe = FileDescriptor(rawValue: 4)
+
         do {
             let spec: ContainerizationOCI.Spec
             do {
@@ -41,7 +44,7 @@ struct RunCommand: ParsableCommand {
             } catch {
                 throw App.Failure(message: "failed to load OCI bundle at \(bundlePath): \(error)")
             }
-            try execInNamespace(spec: spec)
+            try execInNamespace(spec: spec, syncPipe: syncPipe, ackPipe: ackPipe)
         } catch {
             App.writeError(error)
             throw error
@@ -85,8 +88,8 @@ struct RunCommand: ParsableCommand {
 
     private func childSetup(
         spec: ContainerizationOCI.Spec,
-        ackPipe: FileDescriptor,
-        syncPipe: FileDescriptor
+        syncPipe: FileDescriptor,
+        ackPipe: FileDescriptor
     ) throws {
         guard let process = spec.process else {
             throw App.Failure(message: "no process configuration found in runtime spec")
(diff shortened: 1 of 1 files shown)


vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -243,12 +246,44 @@ struct RunCommand: ParsableCommand {
...
+        try errorPipe.close()
+
+        if let linux = spec.linux, !linux.cgroupsPath.isEmpty {
+            let cgroupManager = try Cgroup2Manager.load(group: URL(filePath: linux.cgroupsPath))
+
+            while true {
+                usleep(1_000_000)
+                let events = try cgroupManager.getMemoryEvents()
+
+                if events.max > oomLimit {
```
dkovba, line 276: `events.max` represents the number of events. `1_000_000` seems to be a too large threshold. Would it be appropriate to use a threshold of zero?

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -243,12 +246,44 @@ struct RunCommand: ParsableCommand {
...
+        try syncPipe.close()
+        try ackPipe.close()
+        try errorPipe.close()
+
+        if let linux = spec.linux, !linux.cgroupsPath.isEmpty {
+            let cgroupManager = try Cgroup2Manager.load(group: URL(filePath: linux.cgroupsPath))
+
+            while true {
+                usleep(1_000_000)
+                let events = try cgroupManager.getMemoryEvents()
```
dkovba, line 274: Can we move `MemoryMonitor` to the `Cgroup` library and use it instead of pulling memory events with a fixed interval? CC @dcantah

vminitd/Sources/vmexec/RunCommand.swift
```diff
@@ -243,12 +246,44 @@ struct RunCommand: ParsableCommand {
...
+
+                if events.max > oomLimit {
+                    try cgroupManager.kill()
```
dkovba, line 277: Should we use `App.writeError(error)` and `exit(code)` when `try` fails?

dcantah:
I'm sort of confused on what this is trying to solve. If the idea is we have some oom kills (likely child processes) that happen but init keeps running, there exists a cgroup toggle that makes it such that every process in the cgroup gets killed if there was an oom condition. Meaning, if the init process for the container is well within its limits, but some child process(es) keep getting oom killed, the kernel would kill the whole cgroup (and thus the whole container).

dkovba:
When we run out of memory, a container hangs. The goal is to make it crash with an OOM error.

dcantah:
Ok, regardless of what we decide I don't think we should run a separate forked process to do this. We can try and expose an API on LinuxContainer to monitor memory events in a stream like fashion. If we don't want to to do that either, today this whole scheme could be done with the APIs we expose right now. You could call `LinuxContainer.statistics` every {arbitrary} seconds and check the `memoryEvents` field.

crosbymichael:
The kernel is doing the correct thing here, it is reclaiming memory without SIGKILL. SIGKILL of any process should always be a last resort because of data loss.  We need to look into the root cause for why reclaim is slow.  We don't have swap, therefore in low memory situations, reclaim can take a while and we could have a lock up.  We should look into overhead for the VM and protect responsiveness for vminitd, keeping the container's memory.max < vm memory.

Issue #606, opened 2026-03-30
[Request]: CI: add Linux build and test


jglogan:
### Feature or enhancement request details

Today our build cross compiles for Linux using the Swift static MUSL SDK.

We could better assure that our libraries are Linux compatible by adding a workflow for building and testing using glibc on a Linux runner.

If we do this we should probably also add a Makefile target to perform the same build in a container locally, in case we run into errors in CI.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #654, opened 2026-04-06
Actions: Add Linux build step

dcantah:
Related to https://github.com/apple/containerization/issues/606

To make sure the project (and many linux specific libraries) builds on linux lets add a linux step to build for both musl and glibc. Eventually this will run unit tests as well. This change needs a bit more libc agnostic-ifying work to make sure things work for both libcs.

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -14,7 +14,6 @@ on:
 jobs:
   build:
     name: Linux compile check
-    if: github.repository == 'apple/containerization'
     timeout-minutes: 30
     runs-on: ubuntu-24.04
     container: swift:6.3-noble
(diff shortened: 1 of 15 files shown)


dcantah:
Needs https://github.com/apple/containerization/pull/655

.github/workflows/linux-build.yml
```diff
@@ -0,0 +1,44 @@
...
+    types: [opened, reopened, synchronize]
+  push:
+    branches:
+      - main
+      - release/*
+
+jobs:
+  build:
+    name: Linux compile check
+    if: github.repository == 'apple/containerization'
```
madrob: Probably don't need this since all repos should have the Linux runners available

.github/workflows/linux-build.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Build containerization
+        run: make containerization
+
+      - name: Build vminitd (glibc)
+        run: make -C vminitd SWIFT_CONFIGURATION="--disable-automatic-resolution -Xswiftc -warnings-as-errors"
+
+      - name: Install Static Linux SDK
+        run: |
+          curl -L -o /tmp/static-linux-sdk.tar.gz https://download.swift.org/swift-6.3-release/static-sdk/swift-6.3-RELEASE/swift-6.3-RELEASE_static-linux-0.1.0.artifactbundle.tar.gz
+          swift sdk install /tmp/static-linux-sdk.tar.gz
```
madrob: Check checksums

.github/workflows/linux-build.yml
dcantah: Ack!

.github/workflows/linux-build.yml
```diff
@@ -0,0 +1,40 @@
...
+    branches:
+      - main
+      - release/*
+
+jobs:
+  build:
+    name: Linux compile check
+    timeout-minutes: 30
+    runs-on: ubuntu-24.04
+    container: swift:6.3-noble
```
katiewasnothere, line 19: I would prefer we install swift in the steps so we can programmatically decide what version to install instead of needing to update this container image manually.

katiewasnothere (approved):

PR #655, opened 2026-04-06
vminitd: Make everything libc agnostic

dcantah:
Remove the last (as far as I can see..) of the musl specific imports we had.

### DIFF
--- a/vminitd/Sources/vminitd/PauseCommand.swift
+++ b/vminitd/Sources/vminitd/PauseCommand.swift
@@ -17,7 +17,14 @@
 import ArgumentParser
 import Dispatch
 import Logging
+
+#if canImport(Musl)
 import Musl
+private let _exit = Musl.exit
+#elseif canImport(Glibc)
+import Glibc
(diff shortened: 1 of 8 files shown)


vminitd/Sources/vminitd/PauseCommand.swift
```diff
@@ -17,7 +17,16 @@
 import ArgumentParser
 import Dispatch
 import Logging
+
+#if canImport(Musl)
 import Musl
+private let _exit = Musl.exit
+private let _pause = Musl.pause
```
dkovba: Is an alias actually needed for `pause()`?

dkovba (approved):

vminitd/Sources/vminitd/PauseCommand.swift
dcantah: Good q, let me check. Exit is needed as it conflicts with arg parsers exit

dcantah: Doesn't seem so!

PR #672, opened 2026-04-08
Adds journal modes to `EXT4.Formatter`

jglogan:
**NOTE: Moving to draft as the journal capacity allocation approach in #689 is preferred.**

- Closes #671.
- Adds optional journal parameter to `EXT4.Formatter.init()`, with nil default specifying the current no-journal filesystem configuration. Otherwise the parameter value contains the journal size and mode.


Sources/ContainerizationEXT4/EXT4+Types.swift
```diff
@@ -244,6 +244,15 @@ extension EXT4 {
         public var checksum: UInt32 = 0
     }
 
+    static let JournalMagic: UInt32 = 0xC03B_3998
+    static let JournalInode: InodeNumber = 8
+
+    struct DefaultMountOpts {
+        static let journalData: UInt32 = 0x0004  // data=journal
+        static let journalOrdered: UInt32 = 0x0008  // data=ordered
+        static let journalWriteback: UInt32 = 0x000C  // data=writeback
```
dkovba: The correct values are:
```
#define EXT4_DEFM_JMODE_DATA	0x0020
#define EXT4_DEFM_JMODE_ORDERED	0x0040
#define EXT4_DEFM_JMODE_WBACK	0x0060
```

Sources/ContainerizationEXT4/EXT4+Journal.swift
```diff
@@ -0,0 +1,185 @@
...
+        journalInode.crtimeExtra = now.hi
+        journalInode.linksCount = 1
+        journalInode.extraIsize = UInt16(EXT4.ExtraIsize)
+        journalInode.flags = EXT4.InodeFlag.extents.rawValue
+
+        // Journal is one contiguous allocation → numExtents = 1 → extent tree fits inline
+        // in the inode, so writeExtents needs no extra disk I/O for extent index blocks.
+        // Safe: the journal is placed inside the filesystem, whose total block count is
+        // also a UInt32, so startBlock + blockCount cannot exceed UInt32.max.
+        journalInode = (try? self.writeExtents(journalInode, (startBlock, startBlock + blockCount))) ?? journalInode
```
dkovba: Should we use `try` instead of `try?` here?

Sources/ContainerizationEXT4/EXT4+Journal.swift
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    // MARK: - Private helpers
+
+    private func calculateJournalSize(requestedSize: UInt64?, totalBlocks: UInt32) -> UInt32 {
+        if let size = requestedSize {
+            // Clamp to UInt32.max: the kernel caps internal journals at 2^32 blocks
+            // (per §3.6.4 s_maxlen), and a caller-supplied size large enough to exceed
+            // that would otherwise trap on the narrowing conversion.
+            let blocks = size / UInt64(self.blockSize)
```
dkovba, line 47: Do we need to check that `blocks > 0` here?

Sources/ContainerizationEXT4/EXT4+Types.swift
jglogan: Thanks...turns out the spec I'd written up had the wrong values so these got propagated into the code.

Sources/ContainerizationEXT4/EXT4+Journal.swift
jglogan: the numerator and denominator are both `UInt64` (`self.blockSize` is `UInt32`) so that check would be dead code

dkovba: `block` would be 0 if `size` was less than `self.blockSize`.

dkovba (approved):

Sources/ContainerizationEXT4/EXT4+Journal.swift
```diff
@@ -0,0 +1,185 @@
...
+        let ji = self.inodes[Int(EXT4.JournalInode) - 1].pointee
+        // Extract the 15 UInt32 words from the inode's 60-byte extent-tree field,
+        // then append sizeLow and sizeHigh as words 15 and 16.
+        var words = [UInt32](repeating: 0, count: 17)
+        withUnsafeBytes(of: ji.block) { bytes in
+            for i in 0..<15 {
+                words[i] = bytes.load(fromByteOffset: i * 4, as: UInt32.self)
+            }
+        }
+        words[15] = ji.sizeLow
```
wlan0: The superblock is weird. It doesn't always follow the little endian convention: https://docs.kernel.org/filesystems/ext4/super.html (look for 0x10C)

words[15] and words[16] should be interchanged

wlan0 (changes requested):

Sources/ContainerizationEXT4/EXT4+Journal.swift
jglogan: My brain is fried, I thought I was seeing "do we need to check for negative?"

Sources/ContainerizationEXT4/EXT4+Journal.swift
```diff
@@ -0,0 +1,185 @@
...
+        journalInode.atimeExtra = now.hi
+        journalInode.ctime = now.lo
+        journalInode.ctimeExtra = now.hi
+        journalInode.mtime = now.lo
+        journalInode.mtimeExtra = now.hi
+        journalInode.crtime = now.lo
+        journalInode.crtimeExtra = now.hi
+        journalInode.linksCount = 1
+        journalInode.extraIsize = UInt16(EXT4.ExtraIsize)
+        journalInode.flags = EXT4.InodeFlag.extents.rawValue
```
wlan0: journal Inodes and file Inodes will be of different sizes if `EXT4_HUGE_FILE_FL` is omitted here. Without that flag, the kernel accounts for blocks being 512 bytes in size. However, `writeExtents` works in `blockSize` units. This will cause accounting mismatches. 

The fix is to set `InodeFlag.hugeFile` flag

Sources/ContainerizationEXT4/EXT4+Journal.swift
jglogan: Actually this needs to be bounded at 1024 blocks minimum...

wlan0 (approved):

Sources/ContainerizationEXT4/EXT4+Journal.swift
```diff
@@ -0,0 +1,196 @@
...
+        }
+        let journalStartBlock = self.currentBlock
+        try writeJournalSuperblock(journalBlocks: journalBlocks, filesystemUUID: filesystemUUID)
+        try zeroJournalBlocks(count: journalBlocks - 1)
+        try setupJournalInode(startBlock: journalStartBlock, blockCount: journalBlocks)
+    }
+
+    // MARK: - Private helpers
+
+    private func calculateJournalSize(requestedSize: UInt64?, totalBlocks: UInt32) throws -> UInt32 {
```
wlan0, line 45: JournalSize needs to account for `optimizeBlockGroupLayout`

The totalBlocks when this gets called is an estimate based on blockSize and diskSize. However, based on how much data is written, the layout changes  (right after `close()` gets called. 

The fix would be to either defer the size validation until after `close()` knows the real final block count, or to have `calculateJournalSize` use the post-`optimizeBlockGroupLayout` block count (i.e., account for the fact that close() will expand the image to fit). The `totalBlocks / 2` guard should also be applied consistently to both paths.

Sources/ContainerizationEXT4/EXT4+Formatter.swift
wlan0: This hardcodes `blockSize` - the blockSize argument from the constructor is not going to materially change the value.  

I would remove this value to make the constructor work

wlan0 (changes requested):
https://github.com/apple/containerization/pull/672/changes#r3066127471

Sources/ContainerizationEXT4/EXT4+Formatter.swift
dkovba: We always had `logBlockSize` hardcoded with the meaningless parameter `blockSize`, which had to be exactly `4096` to match the hardcoded `logBlockSize`. We can create an issue for allowing to customize the block size via `logBlockSize` or address this in a separate PR. This is out of scope of this PR.

dkovba: Restored a configurable block size in https://github.com/apple/containerization/pull/679.

jglogan:
Closing, superseded by #689.

PR #676, opened 2026-04-10
Start converting libraries to use FoundationEssentials

dcantah:
This helps on binary size if we can successfully get all Foundation imports out of our dep chain. It seems Foundation brings in a 30MiB ICU blob, which bloats vmexec and vminitd.

### DIFF
--- a/Sources/ContainerizationOS/File.swift
+++ b/Sources/ContainerizationOS/File.swift
@@ -14,7 +14,19 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
-import Foundation
+#if canImport(Musl)
+import Musl
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Darwin)
+import Darwin
+#endif
+
+#if canImport(FoundationEssentials)
+import struct FoundationEssentials.URL
+#else
+import struct Foundation.URL
+#endif
 
 /// Trivial type to discover information about a given file (uid, gid, mode...).
 public struct File: Sendable {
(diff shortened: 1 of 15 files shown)


Sources/ContainerizationOS/Socket/Socket.swift
```diff
@@ -15,6 +15,7 @@
 //===----------------------------------------------------------------------===//
 
 import CShim
+import Dispatch
```
jglogan: Below you're adding Dispatch and removing Foundation, here just adding Dispatch? Is this import used?

Sources/ContainerizationOS/File.swift
```diff
@@ -52,7 +64,7 @@ public struct File: Sendable {
 /// `FileInfo` holds and provides easy access to stat(2) data
 /// for a file.
 public struct FileInfo: Sendable {
-    private let _stat_t: Foundation.stat
+    private let _stat_t: stat
```
jglogan, line 67: What happens here if `!canImport(FoundationEssentials)`?

dcantah: I don't know of a platform where this isn't true. It's available on non-darwin, and on Darwin Foundation is available

dcantah: Oh I see what you mean. On Linux, whatever libc wins and that's where the definition comes from as fessentials doesn't export this. On macOS.. presumably stat is coming from Darwin. Let me see if we even need Foundation here..

Sources/ContainerizationOS/Socket/Socket.swift
dcantah: I think this was probably just a mistake, let me see

dcantah: Ahh I remember, I tried to essentialsify this file, but it uses FileHandle so I can't without a much larger change. I left it as is for now but forgot to remove the Dispatch import

Sources/ContainerizationOS/File.swift
dcantah: I'm going to change to just importing the URL struct

jglogan (approved):

PR #686, opened 2026-04-14
Make unit tests runnable on Linux

### LINKED ISSUE #606: [Request]: CI: add Linux build and test

dcantah:
Closes #606

This lets the unit tests be runnable on linux. This:

- Adds a linux-test makefile target so we can run the unit tests locally
- Fixes up some test code to work on Linux (mostly ifdefs)
- Runs the unit tests in CI now

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -55,3 +55,6 @@ jobs:
 
       - name: Build vminitd (musl)
         run: make -C vminitd
+
+      - name: Run unit tests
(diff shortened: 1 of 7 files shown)


dcantah:
Seems the allocator on the ci machines is being more unfavorable to the ext4 tests

adityaramani (approved):

PR #687, opened 2026-04-14
EXT4: Rework Ptr

dcantah:
Ptr<T> allowed allocation without initialization, which left memory contents undefined in some cases if you didn't call initialize. Reserved inodes (3-10) were allocated but never initialized, causing the dir-count logic to read garbage mode bits depending on the allocator. This came out when trying to get the unit tests to run on an x86 linux box.

This change redesigns Ptr<T> to take a value in init, to circumvent this. All of our uses of Ptr would only allocate memory for 1 item as well, so the capacity field seems unneccessary today. Additionally, most of the methods on the type were unused today so we can clean up quite a few things there too.


wlan0 (approved):
LGTM. This fixes a real correctness issue by removing the allocate-without-initialize from `Ptr`

PR #688, opened 2026-04-16
Epoll: Unit tests

dcantah:
Now that we can run this project on linux, lets add some unit tests for the surfaces that don't..

This additionally changes delete to throw in most cases. I don't really see how masking these errors is ideal.

### DIFF
--- a/Tests/ContainerizationOSTests/EpollTests.swift
+++ b/Tests/ContainerizationOSTests/EpollTests.swift
@@ -1,0 +1,300 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the Containerization project authors.
+//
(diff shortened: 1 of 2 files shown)


Tests/ContainerizationOSTests/EpollTests.swift
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    @Test
+    func deletingAlreadyClosedFDThrows() throws {
+        let epoll = try Epoll()
+        let (readFD, writeFD) = try Self.makePipe()
+        try epoll.add(readFD, mask: .input)
+        close(readFD)
+        close(writeFD)
```
realrajaryan, line 109: `defer`?

realrajaryan (approved):

PR #690, opened 2026-04-18
Add support for network block device (NBD)

yibozhuang:
This change adds support for attaching network block device (NBD) to both LinuxContainer and LinuxPod.

For LinuxContainer, whether to use the underlying `VZNetworkBlockDeviceStorageDeviceAttachment` is determined by the URL of the container Mount source.

For LinuxPod, adds additional API to support pod-level volumes that can be mounted into multiple containers. The PodVolume type provides enum to support multiple types of volume source. LinuxContainer can reference the pod level volume using the `Mount.sharedMount()` constructor with the name referencing the name of the pod volume. This will allow the NBD to be attached to the pod at the VM level and then bind-mounted into the container.

For integration tests, added a lightweight NBD server implementation in swift that speaks the NBD protocol to ensure there is sufficient coverage for the changes introduced.

Apple Virtualization NBD support documentation:
https://developer.apple.com/documentation/virtualization/vznetworkblockdevicestoragedeviceattachment

### DIFF
--- a/Sources/Containerization/LinuxPod.swift
+++ b/Sources/Containerization/LinuxPod.swift
@@ -59,6 +59,8 @@ public final class LinuxPod: Sendable {
         /// The default hosts file configuration for all containers in the pod.
         /// Individual containers can override this by setting their own `hosts` configuration.
         public var hosts: Hosts?
+        /// Volumes attached to the pod. Can be shared with multiple containers.
+        public var volumes: [PodVolume] = []
 
         public init() {}
     }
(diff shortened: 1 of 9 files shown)


Sources/Integration/PodTests.swift
```diff
@@ -2032,4 +2032,61 @@ extension IntegrationSuite {
             throw error
         }
     }
+
```
dcantah, line 2035: Can we have LinuxContainer tests?

yibozhuang: It is added here: https://github.com/apple/containerization/pull/690/changes#diff-894a721a2c7fad8840edb5e8b3af746406b947a53d5f0764c8d7fcfe13c2def7R63

Sources/Containerization/LinuxPod.swift
```diff
@@ -86,10 +88,70 @@ public final class LinuxPod: Sendable {
         /// Run the container with a minimal init process that handles signal
         /// forwarding and zombie reaping.
         public var useInit: Bool = false
+        /// The container mounts that references pod-level attached volumes.
+        public var volumeMounts: [VolumeMount] = []
```
dcantah: I'm not super happy with this in particular but I can't put my finger on it. This almost seems like a UX win really right? There's nothing they couldn't emulate by just asking for the same mount in every container in the pod?

dcantah: I think if we were to keep this I'd prefer sharedMounts or something.

yibozhuang: @dcantah How about this 

- remove the `volumeMounts` from `ContainerConfiguration`
- Introduce a new `Mount.sharedMount()` constructor to indicate that this is a shared mount
- Add a `.shared` RuntimeOptions to indicate that this is a shared mount between containers
- `PodContainer` can do `config.mounts.append(Mount.sharedMount(name: "shared-data", destination: "/data", options: ["ro"]))` to add a shared mount to its configuration, no additional API surface needed

Does this work better?

yibozhuang: Updated code to use ^ implementation.

dcantah (approved):

PR #692, opened 2026-04-20
Terminal: Rework type

dcantah:
To try and purge Foundation from the Linux builds I've been making the rounds on our types. This one is a breaking change, but I'd been wanting to do this for ages. `Terminal` (to me) should be as lightweight as possible and not be tied to any `FileHandle` semantics. I think we should expose the raw fd, but exposing a handle parameter was always a bit odd to me. This change:

- Gets rid of the `handle` var.
- Keeps `Terminal` a value type, so if you did want some double close protection or synchronization on the type you can decide to wrap it if you need to.
- Exposes a `fileDescriptor` public var.
- Adds unit tests for Terminal finally..

### DIFF
--- a/Sources/ContainerizationOS/Terminal.swift
+++ b/Sources/ContainerizationOS/Terminal.swift
@@ -14,16 +14,31 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
 import Foundation
+#endif
+
+#if canImport(Musl)
+import Musl
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Darwin)
+import Darwin
+#else
+#error("Unsupported platform")
+#endif
+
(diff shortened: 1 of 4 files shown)


Sources/ContainerizationOS/Terminal.swift
```diff
@@ -14,29 +14,57 @@
...
             initState = nil
         }
-        self.handle = .init(fileDescriptor: descriptor, closeOnDealloc: false)
+        self.fileDescriptor = descriptor
     }
 
     /// Write the provided data to the tty device.
     public func write(_ data: Data) throws {
-        try handle.write(contentsOf: data)
+        try data.withUnsafeBytes { buffer in
```
crosbymichael, line 54: If we are going to be moving away from file handle we should probably have some extensions on data or the file descriptor to write() like you have done here, handling retry and such.

dcantah: I'm curious if FileDescriptor from System already does this for us

crosbymichael:
We can address then when you rebase or want to continue the work :)

PR #693, opened 2026-04-20
Refactor ArchiveWriter to expose archiving each URL independently

JaewonHur:
This PR refactors `ArchiveWriter` to add an API `archive(_ paths: base:)`. This API is used to archive the contents at each URL independently, similar to doing `tar -cvf archive.tar foo.bin /bar/baz.txt`.

### DIFF
--- a/Sources/ContainerizationArchive/ArchiveWriter.swift
+++ b/Sources/ContainerizationArchive/ArchiveWriter.swift
@@ -179,6 +179,94 @@ extension ArchiveWriter {
 }
 
 extension ArchiveWriter {
+    private func archive(_ relativePath: FilePath, dirPath: FilePath) throws {
+        let fm = FileManager.default
+
+        let fullPath = dirPath.appending(relativePath.string)
+
+        var statInfo = stat()
+        guard lstat(fullPath.string, &statInfo) == 0 else {
+            let errNo = errno
+            let err = POSIXErrorCode(rawValue: errNo) ?? .EINVAL
+            throw ArchiveError.failedToCreateArchive("lstat failed for '\(fullPath)': \(POSIXError(err))")
+        }
+
+        let mode = statInfo.st_mode
+        let uid = statInfo.st_uid
+        let gid = statInfo.st_gid
+        var size: Int64 = 0
+        let type: URLFileResourceType
+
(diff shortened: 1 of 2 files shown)


Sources/ContainerizationArchive/ArchiveWriter.swift
```diff
@@ -214,88 +302,32 @@ extension ArchiveWriter {
...
+        }
+    }
 
-            var statInfo = stat()
-            guard lstat(fullPath.string, &statInfo) == 0 else {
-                let errNo = errno
-                let err = POSIXErrorCode(rawValue: errNo) ?? .EINVAL
-                throw ArchiveError.failedToCreateArchive("lstat failed for '\(fullPath)': \(POSIXError(err))")
-            }
+    public func archiveURLs(_ urls: [URL], base: URL) throws {
```
dcantah: I really do not want the path type (URL) to be in the name. Maybe `archivePaths`, or even just `archive`? We can rely on overloading to support an array of `FilePath`s in the future for either version.

jglogan:
@JaewonHur How much overlap is there with #652?

Sources/ContainerizationArchive/ArchiveWriter.swift
jglogan: `archive(_ paths: [FilePath], base: FilePath)` gets my vote.

For container 1.0 we'll be banishing URL wherever possible.

dcantah: I'm okay with going to FilePath here to start.

JaewonHur: Updated to use `archive(_ paths: [FilePath], base: FilePath)`.
It feels somewhat weird as `archive(_ paths: [FilePath],...)` needs full paths, while `archive(_ relativePath: FilePath,...)` needs a relative path, but leave for now.

I can surely update if you want!

egernst:
Should we rebase this one?

dcantah:
You need to sign your commits. The change lgtm at this point after. Just squash everything down to 1 commit

egernst (approved):
Thanks for addressing the feedback.

PR #698, opened 2026-04-23
LinuxContainer/LinuxPod: Make time sync configurable

dcantah:
Today our time syncing with the host lives in VZVirtualMachineInstance, which means if a user made their own vmm type they would have to roll their own time syncing. It likely makes more sense to hoist this up to LinuxContainer/LinuxPod so if they want it they can use it, and if they have their own synchronization they can just disable it and use whatever they want. This also lowers the default sync period to 20 seconds.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -78,6 +78,9 @@ public final class LinuxContainer: Container, Sendable {
         /// Run the container with a minimal init process that handles signal
         /// forwarding and zombie reaping.
         public var useInit: Bool = false
+        /// Interval for syncing the guest clock with the host. Set to nil
+        /// to disable time synchronization.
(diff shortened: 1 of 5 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -78,6 +78,9 @@ public final class LinuxContainer: Container, Sendable {
         /// Run the container with a minimal init process that handles signal
         /// forwarding and zombie reaping.
         public var useInit: Bool = false
+        /// Interval for syncing the guest clock with the host. Set to nil
+        /// to disable time synchronization.
+        public var timeSyncInterval: Duration? = .seconds(20)
```
jglogan, line 83: Is `timeSyncInterval` a property of every time sync implementation that could possibly exist? Or could we just inject a protocol-conforming TimeSyncer that defines a policy for calling `agent.setTime()`?

PR #685, opened 2026-04-14
[vminitd]: api for freeze/thaw filesystem operations

saehejkang:
Addition of `vminitd` API for freeze/thaw filesystem operations

Closes #660

PR #700, opened 2026-04-27
[vminitd]: api for trim filesystem operations

### LINKED ISSUE #773: [Request]: Add trim to FilesystemOperation API in vminit.

Once the filesystem operation API has been added for freezing and thawing a filesystem, having the ability to trim blocks (FITRIM ioctl) would be useful for returning unused blocks to the host.

### Code of Conduct

saehejkang:
Addition of `vminitd` API for trim filesystem operations

Closes #773

Depends on #685

### DIFF
--- a/Sources/Containerization/SandboxContext/SandboxContext.proto
+++ b/Sources/Containerization/SandboxContext/SandboxContext.proto
@@ -30,6 +30,8 @@ service SandboxContext {
   rpc Copy(CopyRequest) returns (stream CopyResponse);
   // Stat a path in the guest filesystem.
   rpc Stat(StatRequest) returns (StatResponse);
+  // Perform a filesystem operation on a mounted filesystem.
+  rpc FilesystemOperation(FilesystemOperationRequest) returns (FilesystemOperationResponse);
 
   // Create a new process inside the container.
   rpc CreateProcess(CreateProcessRequest) returns (CreateProcessResponse);
@@ -291,6 +293,19 @@ message StatResponse {
   Stat stat = 1;
   string error = 2; // Non-empty if stat failed.
 }
+
+message FiFreezeParams {}
+message FiThawParams {}
+
+message FilesystemOperationRequest {
+  string path = 1;
+  oneof operation {
+    FiFreezeParams freeze = 2;
+    FiThawParams thaw = 3;
+  }
+}
+
+message FilesystemOperationResponse {}
 
 message IpLinkSetRequest {
   string interface = 1;
--- a/Sources/Containerization/SandboxContext/SandboxContext.proto
+++ b/Sources/Containerization/SandboxContext/SandboxContext.proto
@@ -294,18 +294,34 @@ message StatResponse {
   string error = 2; // Non-empty if stat failed.
 }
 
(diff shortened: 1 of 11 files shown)


saehejkang:
Kind of messy, as the work is added to the freeze/thaw filesystem PR, but ready for initial review

Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -292,6 +294,33 @@ message StatResponse {
   string error = 2; // Non-empty if stat failed.
 }
 
+message FiTrimParams {
```
jglogan, line 297: For FITRIM I don't think we want a 1:1 mapping between the ioctl parameters and the vminit API parameters. At least for now, we want to trim the entire volume that path points to:

```c
struct fstrim_range range = {
    .start = 0,
    .len = ULLONG_MAX,
    .minlen = 0,
};
```

jglogan: What we want to build toward for this call is something where we define trim policies. For this PR, `schedule` could just contain OneShot, and we could add Recurring in a follow-up PR.

Once the one-shot operation is merged, you can implement a `container clean` command that runs one shot trim on the container root fs path, and the mount paths for each named volume.

Recurring would trim the named path periodically, with a duration of zero clearing a previously established schedule for the path.

Discussion topics once we get to the recurring trim policy include:

- Should the call naively trim `path`, or should it infer the mount point such that callers don't inadvertently install multiple periodic trims that operate on the same mount?
- Should there be a Params value that clears all previously established trim policies?

```grace
import "google/protobuf/duration.proto";

message FiTrimParams {
  oneof schedule {
    OneShot one_shot = 1;
    Recurring recurring = 2;
  }

  message OneShot {}

  message Recurring {
    google.protobuf.Duration interval = 1;
  }
}
```

saehejkang: Will come back to this comment/PR when https://github.com/apple/containerization/pull/685 is completed and merged.

Sources/Containerization/VirtualMachineAgent.swift
```diff
@@ -28,6 +28,7 @@ public struct WriteFileFlags {
 public enum FilesystemOperation: Sendable {
     case freeze
     case thaw
+    case trim
```
jglogan, line 31: Let's ignore the return value for the trim operation. It's the number of bytes that the OS asks the storage device to trim, but it's not necessarily what's discarded.

If we did care about the value, `filesystemOperation()` shouldn't have a `UInt64?` return value. It should instead return a result enum; for example `Filesystem.Result`:

```swift
  public enum FilesystemOperation: Sendable {
      case freeze
      case thaw
      case trim

      public enum Result: Sendable {
          case freeze
          case thaw
          case trim(bytes: UInt64)
      }
  }
```

saehejkang: Got it

jglogan (approved):

jglogan (approved):

PR #704, opened 2026-04-28
Rework signals usage/parsing

dcantah:
Replace the old platform-gated Signals helper in ContainerizationOS with a Signal struct in Containerization. Signal provides named constants for common signals (.kill, .term, etc.), platform-specific namespaces (Signal.Linux for rtmin/stkflt/pwr, Signal.Darwin for emt/info), string parsing, and ExpressibleByIntegerLiteral support.

LinuxContainer, LinuxProcess, and LinuxPod kill APIs now take Signal instead of raw Int32.


crosbymichael (approved):
Very nice

PR #708, opened 2026-04-28
make vminitd package for extensions

crosbymichael:


vminitd/Sources/VminitdCore/CommandRunner.swift
```diff
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2025-2026 Apple Inc. and the Containerization project authors.
+// Copyright © 2026 Apple Inc. and the Containerization project authors.
```
dkovba, line 2: Consider reverting these changes.

vminitd/Sources/vminitd/AgentCommand.swift
```diff
@@ -81,12 +80,6 @@ struct AgentCommand: AsyncParsableCommand {
         }
         log.info("vminitd booting", metadata: metadata)
 
-        // Set of mounts necessary to be mounted prior to taking any RPCs.
-        // 1. /proc as the sysctl rpc wouldn't make sense if it wasn't there (NOTE: This is done before this method
-        // due to Swift seemingly requiring /proc to be present for the async runtime to spin up).
-        // 2. /run as that is where we store container state.
-        // 3. /sys as we need it for /sys/fs/cgroup
-        // 4. /sys/fs/cgroup to add the agent to a cgroup, as well as containers later.
```
dcantah, line 89: Why'd we lose the comments?

vminitd/Sources/VminitdCore/CommandRunner.swift
crosbymichael: `make fmt` changes these because the files are not modified. This is correct.

dcantah (approved):
The public blast radius isn't as much as I thought which is nice

saehejkang:
Thank you for this! Solved the indexing issues I was having.

PR #709, opened 2026-04-28
Add bridge mode networking

### LINKED ISSUE #457: [Request]: Add support for Bridged Networking

It would be very useful is this library would support bridged networking, i.e. giving every VM/container their own IP on the LAN. We could re-use the mac address parameter for that.

In the container package, a network could have a type `nat` or `bridged`, I think.

### Code of Conduct

torarnv:
Fixes #457

### DIFF
--- a/Sources/Containerization/BridgedNetworkInterface.swift
+++ b/Sources/Containerization/BridgedNetworkInterface.swift
@@ -1,0 +1,59 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the Containerization project authors.
+//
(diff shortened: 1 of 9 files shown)


torarnv:
Early RFC. See also https://github.com/apple/container/pull/1463

dcantah:
Some quick fixes so we can get some CI signal:

1. We require all commits be signed (ssh/gpg)
2. Run `make fmt` and the formatting yell from our CI should go poof

torarnv:
> 1. We require all commits be signed (ssh/gpg)
> 2. Run `make fmt` and the formatting yell from our CI should go poof

Done 👍🏻

PR #713, opened 2026-04-29
Fix head-of-line blocking and resource leak in BidirectionalRelay

### LINKED ISSUE #712: [Bug]: Head-of-line blocking in BidirectionalRelay causes permanent hang under concurrent vsock proxy load
### I have done the following

### Steps to reproduce

1. Run an application that opens many simultaneous socket connections through the `UnixSocketRelay` vsock proxy (e.g., multiple concurrent operations that stress the proxy heavily)
2. Generate enough concurrent traffic that at least one destination file descriptor's write buffer fills up
3. Attempt any new operation through the proxy (e.g., a simple status check)

The hang is reliably reproducible by launching 5+ concurrent container operations simultaneously. Once triggered, the proxy never recovers — all existing and new connections hang indefinitely until the process is restarted.

### Current behavior

The entire vsock proxy becomes permanently unresponsive. All connections — including completely unrelated new ones — hang indefinitely. The process must be fully restarted to recover.

**Root cause:** `BidirectionalRelay` uses synchronous `write()` calls on a shared serial `DispatchQueue` (`com.apple.containerization.socket-relay`). When a destination fd's buffer fills (remote end temporarily busy), the `write()` blocks the serial queue. Since all relay connections share this queue, a single blocked write freezes every connection.

**Secondary issue:** When a relay is torn down while under backpressure (read source suspended), `stop()` calls `cancel()` on the suspended `DispatchSourceRead`. GCD does not deliver cancel handlers on suspended dispatch sources, so file descriptors leak and `waitForCompletion()` never resolves.

### Expected behavior

When one connection's destination buffer is full, only that connection should stall. All other relay connections should continue operating normally. When backpressure clears, the stalled connection should resume without intervention.

### Environment

```markdown
- OS: macOS 26.x (Tahoe)
- Xcode: 26.4
- Swift: 6.3
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

mlsad3:
Replace synchronous `write()` calls in `BidirectionalRelay` with non-blocking I/O and `DispatchSourceWrite` backpressure handling. Under concurrent vsock proxy load, a single blocked write on the shared serial dispatch queue would freeze all relay connections permanently, including unrelated new connections.

Closes #712

### Changes

- Set relay file descriptors to `O_NONBLOCK` and handle `EAGAIN` by suspending reads and installing a write source to drain pending data
- Give each `BidirectionalRelay` its own serial queue instead of sharing one from `UnixSocketRelayManager`, eliminating cross-connection blocking
- Resume suspended read sources before cancelling in `stop()` — GCD does not deliver cancel handlers on suspended dispatch sources, which caused file descriptor and memory leaks on teardown under backpressure
- Guard-unwrap `buf.baseAddress` in `drainPendingWrite`

### Test plan

Added `BidirectionalRelayTests` with 5 tests:
- Basic bidirectional relay
- Cross-connection head-of-line blocking isolation
- Backpressure recovery (`EAGAIN` → write source drain → resume)
- EOF handling
- Stop while backpressured (verifies cancel handlers fire and `waitForCompletion()` resolves)

Tested under sustained concurrent load that previously caused permanent hangs; verified normal operations work correctly after sustained load.

### DIFF
--- a/Tests/ContainerizationOSTests/BidirectionalRelayTests.swift
+++ b/Tests/ContainerizationOSTests/BidirectionalRelayTests.swift
@@ -284,4 +284,56 @@ final class BidirectionalRelayTests {
 
         #expect(completed, "Relay should complete after both sides reach EOF")
     }
+
+    // MARK: - Test 5: Stop while under backpressure
+
+    @Test
+    func testStopWhileBackpressured() async throws {
+        // Verify that stop() works correctly when a read source is suspended
+        // due to backpressure. Previously, cancelling a suspended dispatch source
+        // would never deliver the cancel handler, leaking file descriptors.
+        let (a0, a1) = try makeSocketPair()
+        let (b0, b1) = try makeSocketPair()
+
+        // Shrink b0's send buffer so backpressure kicks in quickly.
+        setSendBufferSize(fd: b0, size: 4096)
+
+        let relay = BidirectionalRelay(fd1: a1, fd2: b0)
+        try relay.start()
+
+        // Write enough to trigger backpressure but don't read from b1.
+        let largeData = [UInt8](repeating: 0x42, count: 65536)
+        let a0flags = fcntl(a0, F_GETFL)
+        _ = fcntl(a0, F_SETFL, a0flags | O_NONBLOCK)
+        _ = largeData.withUnsafeBufferPointer { buf in
+            write(a0, buf.baseAddress!, buf.count)
+        }
+
+        // Give relay time to enter backpressure state (readSource suspended).
+        usleep(100_000) // 100ms
+
+        // Stop the relay while it's backpressured. This should not hang or leak.
+        relay.stop()
+
+        // Close the external ends — the relay's fds should already be closed by stop().
+        close(a0)
+        close(b1)
+
+        // The relay should complete (cancel handlers should have fired).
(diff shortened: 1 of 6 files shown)


Tests/ContainerizationOSTests/BidirectionalRelayTests.swift
```diff
@@ -0,0 +1,341 @@
...
+import Musl
+#endif
+
+@Suite("BidirectionalRelay tests")
+final class BidirectionalRelayTests {
+
+    /// Creates a Unix domain socket pair and returns (fd0, fd1).
+    private func makeSocketPair() throws -> (Int32, Int32) {
+        var fds: [Int32] = [0, 0]
+        let result = socketpair(AF_UNIX, SOCK_STREAM, 0, &fds)
```
jglogan, line 37:
```suggestion
        #if canImport(Darwin)                                                                                                                                                                                                                               
        let result = socketpair(AF_UNIX, SOCK_STREAM, 0, &fds)                                                                                                                                                                 
        #else                                                                                                                                                                                                                                               
        let result = socketpair(AF_UNIX, Int32(SOCK_STREAM.rawValue), 0, &fds)   
        #endif                                                                                                                                   
```

jglogan (approved):

PR #716, opened 2026-05-01
Fix ArchiveReader to support readonly directory

JaewonHur:
Defer attribute setting of directories so that we can unarchive child entries first, then apply attributes which might be read-only.


crosbymichael:
Looks good, can you rebase?

JaewonHur:
@crosbymichael rebased.

PR #719, opened 2026-05-01
TerminalIO: Fix ebadf on close

dcantah:
The stdout and stdin IOPairs in TerminalIO share the same Terminal fd. Closing stdin first would close the Terminal, causing stdout to read -1 from fileDescriptor and fail to unregister from epoll. This left stale handler closures (capturing the IOPair) in ProcessSupervisor.handlers indefinitely.

Additionally, vsock sockets created in TerminalIO.start() with closeOnDeinit disabled were never closed if the process exited before attach() created IOPairs.

### DIFF
--- a/vminitd/Sources/VminitdCore/IOCloser.swift
+++ b/vminitd/Sources/VminitdCore/IOCloser.swift
@@ -19,3 +19,15 @@ protocol IOCloser: Sendable {
 
     func close() throws
 }
+
+struct UnownedIOCloser: IOCloser {
+    private let inner: IOCloser
+
+    var fileDescriptor: Int32 { inner.fileDescriptor }
+
+    init(_ inner: IOCloser) {
+        self.inner = inner
+    }
+
+    func close() throws {}
+}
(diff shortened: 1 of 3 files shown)


vminitd/Sources/VminitdCore/IOCloser.swift
```diff
@@ -19,3 +19,15 @@ protocol IOCloser: Sendable {
 
     func close() throws
 }
+
```
jglogan, line 22: docc for this protocol and the conforming types would be nice

jglogan: but not today

vminitd/Sources/VminitdCore/TerminalIO.swift
```diff
@@ -98,7 +98,7 @@ final class TerminalIO: ManagedProcess.IO & Sendable {
             if let stdinSocket = $0.stdinSocket {
                 let pair = IOPair(
                     readFrom: stdinSocket,
-                    writeTo: term,
+                    writeTo: UnownedIOCloser(term),
```
jglogan, line 101: comment about why UnownedIOCloser would be useful

jglogan (approved):

PR #724, opened 2026-05-04
Support running native process outside sandbox

JaewonHur:
This PR supports `NativeProcess`, which runs outside of containerized environment. `NativeProcess` is managed the same as `ManagedProcess`, except it is spawned directly by `vminitd` (not `vexec`). These processes run outside of container namespaces, and no resources shared with container, invisible to it.

### DIFF
--- a/Sources/Containerization/CreateProcessOptions.swift
+++ b/Sources/Containerization/CreateProcessOptions.swift
@@ -1,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+public struct CreateProcessOptions: Sendable, Codable {
+    /// The process is created outside containerized env.
+    public var native: Bool
+
+    enum CodingKeys: String, CodingKey {
+        case native
+    }
+
+    public init(native: Bool) {
+        self.native = native
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        native = try container.decodeIfPresent(Bool.self, forKey: .native) ?? false
+    }
+}
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -869,6 +869,7 @@ extension LinuxContainer {
     public func kill(_ signal: Signal) async throws {
         try await self.state.withLock {
             let state = try $0.startedState("kill")
+
             try await state.process.kill(signal)
         }
     }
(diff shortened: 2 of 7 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -849,6 +849,11 @@ extension LinuxContainer {
     public func kill(_ signal: Signal) async throws {
         try await self.state.withLock {
             let state = try $0.startedState("kill")
+
+            for process in state.vendedProcesses.values where process.native {
+                try await process.kill(signal)
+            }
```
dcantah: Why can't they call `kill` on the LinuxProcess they get back from exec?

Sources/Containerization/LinuxContainer.swift
```diff
@@ -916,7 +921,7 @@ extension LinuxContainer {
 
     /// Execute a new process in the container. The process is not started after this call, and must be manually started
     /// via the `start` method.
-    public func exec(_ id: String, configuration: LinuxProcessConfiguration) async throws -> LinuxProcess {
+    public func exec(_ id: String, configuration: LinuxProcessConfiguration, native: Bool = false) async throws -> LinuxProcess {
```
dcantah, line 940: Documentation on what native means and does

Sources/Containerization/CreateProcessOptions.swift
```diff
@@ -0,0 +1,34 @@
...
+    public init(native: Bool) {
+        self.native = native
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        native = try container.decodeIfPresent(Bool.self, forKey: .native) ?? false
+    }
+}
```
dcantah, line 34: The createProcess API takes in an optional containerID. This is what I was going to use to signify that we'd want to run a process in the root of the VM. e.g. supplying a nil containerID.

dcantah: I'd prefer we don't introduce new public API like this without really thinking through it

Sources/Containerization/LinuxContainer.swift
JaewonHur: Based on the `SandboxService.stop` source code, I assumed it's the responsibility of `LinuxContainer` to clean up processes in VM. `SandboxService.stop` doesn't `kill` each process, but just `kill` container. Then, `LinuxContainer` kills the init process only, and it's up to the init process whether or not to kill other processes as well (as all others are parented to init process).

But, it's different for `native process` as it is not under container init process, but directly the child of `vminitd`. So, I guess for `LinuxContainer.kill` to be self-confined, it should kill `native process` as well.

My guess was:
* Container's init process is responsible for other processes in container (that's why `LinuxContainer.kill` only kills that init process).
* `vminitd` is responsible for that init process, and also `native processes` (that's why I think `LinuxContainer.kill` should kill `native processes` as well.

Sources/Containerization/CreateProcessOptions.swift
JaewonHur: Oh I forgot. Yeah let me update.
Do you think `NativeProcess` should live outside of `ManagedContainer`?

This will need reimplementing bookkeeping structure and management logic for `NativeProcess`.

Sources/Containerization/LinuxContainer.swift
dcantah: The thing is, you can send ANY signal in kill. The signal doesn't always lead to the container dying. You can send sigwinch if you wanted, which delivering that signal to any native processes might be wrong.

Sources/Containerization/CreateProcessOptions.swift
dcantah: > Do you think NativeProcess should live outside of ManagedContainer?

Whatever you think is right. If it feels too much hassle let me know

PR #725, opened 2026-05-05
refactor memory alignment to VZVirtualMachineInstance

yibozhuang:
Move the MiB rounding from LinuxContainer into `VZVirtualMachineInstance`, so all callers are covered in one place and the alignment constraint stays with the VMM layer that owns it.

This will also ensure LinuxPod can get memory alignment for free.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -551,8 +551,7 @@ extension LinuxContainer {
             var modifiedRootfs = self.rootfs
             modifiedRootfs.options.removeAll(where: { $0 == "ro" })
 
-            let mib: UInt64 = 1.mib()
-            let vmMemory = (self.memoryInBytes + self.config.memoryOverhead + mib - 1) & ~(mib - 1)
+            let vmMemory = self.memoryInBytes + self.config.memoryOverhead
 
             let vmCpus = self.cpus + self.config.cpuOverhead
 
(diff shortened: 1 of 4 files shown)


Sources/Containerization/LinuxPod.swift
```diff
@@ -412,9 +412,12 @@ extension LinuxPod {
                 mountsByID[self.id] = podVolumeMounts
             }
 
+            let mib: UInt64 = 1 << 20
+            let vmMemory = (self.config.memoryInBytes + mib - 1) & ~(mib - 1)
```
dcantah: I'm wondering if we just put this in VZVirtualMachineInstance, although maybe there's other VMMs that have this restriction also.

dcantah (approved):

Sources/Containerization/LinuxPod.swift
yibozhuang: cloud-hypervisor requires page size alignment
https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/vmm/src/memory_manager.rs#L646

qemu looks like will automatically align to 8 KiB
https://github.com/qemu/qemu/blob/master/hw/core/machine.c#L586

So right now if we set `config.memoryInBytes` to `1_000_000_000` then it would fail if VMM is cloud-hypervisor. 

kata-containers does MiB alignment by rounding down: https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/resource/src/cpu_mem/mem.rs#L92

yibozhuang: Updated this PR to move the memory alignment code into `VZVirtualMachineInstance` which I think is the better abstraction and not need to have VMM specific requirement in `LinuxContainer` or `LinuxPod` itself.

dcantah (approved):

PR #731, opened 2026-05-10
vminitd: Fix UDS length issue with large container IDs

dcantah:
Today we craft the proxied sockets we create at a path that contains the container ID. This is normally completely fine, except if the ID is enormous, as it can cause the UDS path to be longer than the allowed length.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -447,8 +447,8 @@ public final class LinuxContainer: Container, Sendable {
         "/run/container/\(id)/rootfs"
     }
 
-    private static func guestSocketStagingPath(_ containerID: String, socketID: String) -> String {
-        "/run/container/\(containerID)/sockets/\(socketID).sock"
+    private static func guestSocketStagingPath(_ socketID: String) -> String {
+        "/run/sockets/\(socketID).sock"
     }
 }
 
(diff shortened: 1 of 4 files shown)


yibozhuang (approved):

Sources/Containerization/LinuxContainer.swift
```diff
@@ -447,8 +447,8 @@ public final class LinuxContainer: Container, Sendable {
         "/run/container/\(id)/rootfs"
     }
 
-    private static func guestSocketStagingPath(_ containerID: String, socketID: String) -> String {
-        "/run/container/\(containerID)/sockets/\(socketID).sock"
+    private static func guestSocketStagingPath(_ socketID: String) -> String {
+        "/run/container/sockets/\(socketID).sock"
```
JaewonHur: What if user creates a container with id `sockets`?

dcantah: Great question.. I think things would still work as we wouldn't make any files named {uuid}.sock, but honestly we should put these at /run/sockets probably.

dcantah: Well actually, the directory will get removed when the container gets deleted so this is a legitimate concern. Moving to /run/sockets

dcantah: Done!

yibozhuang (approved):

PR #733, opened 2026-05-10
LinuxContainer/LinuxPod: Add ID length restriction

dcantah:
I think realistically we should have a limit on this. Technically today the ID length restriction is whatever our grpc buffer length is..

I think 64 is long enough so that folks could use an entire hex based sha256, any uuid, and also sethostname has a length restriction of 64 anyways.


yibozhuang (approved):

JaewonHur (approved):

fredrikaverpil:
@dcantah Hi, quick question; could this change block container names with >64 characters, like the ones created by testcontainers here?

https://github.com/testcontainers/testcontainers-go/blob/982fb09da04d2a525ac55ed0f9634d2111946455/reaper.go#L84-L88

PR #739, opened 2026-05-14
Add LinuxContainer block I/O resources

chrisgeo:
Adds ability to expose blockio flags in support of: [https://github.com/apple/container/issues/1512](apple/container#1512)

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -415,7 +415,7 @@ public final class LinuxContainer: Container, Sendable {
                 quota: Int64(config.cpus * 100_000),
                 period: 100_000
             ),
-            blockIO: config.blockIO
+            blockIO: config.blockIO?.toOCI()
         )
 
         spec.linux?.namespaces = [
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -57,6 +57,8 @@ public final class LinuxContainer: Container, Sendable {
         public var cpus: Int = 4
         /// The memory in bytes to give to the container.
         public var memoryInBytes: UInt64 = 1024.mib()
+        /// Optional block I/O resource limits for the container cgroup.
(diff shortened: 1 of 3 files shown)


Sources/Containerization/LinuxContainer.swift
```diff
@@ -56,6 +56,8 @@ public final class LinuxContainer: Container, Sendable {
         public var cpus: Int = 4
         /// The memory in bytes to give to the container.
         public var memoryInBytes: UInt64 = 1024.mib()
+        /// Optional block I/O resource limits for the container cgroup.
+        public var blockIO: LinuxBlockIO?
```
dcantah, line 61: I'd prefer we don't use the OCI types directly (it gives us a little wiggle room to add various knobs). You can add a new LinuxBlockIO type in Containerization and use this and just convert to the oci variant like we do for other things in this file.

chrisgeo: Thanks for the feedback.

Does this new [shape](https://github.com/apple/containerization/pull/739/changes/5268c48d6a82c5a61e7a73da4c9b3a165cdd8f53..3d009dfc2724fc924dd18ca2014b5e839d554163#diff-a998e035f421f4959805b7945fb79d598e6cd43ebc2a74d1b807aa2b17df4738R20) align with what you were imagining?

jglogan: @dcantah @chrisgeo see https://github.com/apple/container/issues/1512#issuecomment-4480580311 for thoughts on `container` UX as well.

dcantah: Sorry for the radio silence here, I'm trying to reason with what is the best way to go about this. Today we don't expose any cgroup toggles, but the cpu and memory fields we have indirectly set them, so most of it is abstracted. I think we likely will want to have a top level `CgroupOptions` struct that houses these blockio options, and which allows us to add future cgroup toggles. I'll try and have an answer today or tomorrow. We really appreciate you working on this

PR #742, opened 2026-05-18
Move the `VminitdCore` library to the top-level package

dkovba:
External Swift Package Manager consumers can only see products declared in the top-level `Package.swift`. The `VminitdCore` library was declared inside the `vminitd` subpackage, so it was impossible to build a custom `vminitd` against it via a URL dependency on `containerization`. This change moves the `VminitdCore` to the top-level `Package.swift`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -37,7 +37,7 @@ let package = Package(
         .library(name: "ContainerizationOS", targets: ["ContainerizationOS"]),
         .library(name: "ContainerizationExtras", targets: ["ContainerizationExtras"]),
         .library(name: "ContainerizationArchive", targets: ["ContainerizationArchive"]),
-        .library(name: "VminitdCore", targets: ["VminitdCore", "Cgroup", "LCShim"]),
+        .library(name: "VminitdCore", targets: ["VminitdCore", "LCShim"]),
         .executable(name: "cctl", targets: ["cctl"]),
     ],
     dependencies: [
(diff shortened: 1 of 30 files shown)


Package.swift
```diff
@@ -259,6 +264,50 @@ let package = Package(
...
+                .define("GIT_TAG", to: "\"\(gitTag)\""),
+                .define("BUILD_TIME", to: "\"\(buildTime)\""),
+            ]
+        ),
+        .target(
+            name: "LCShim",
+            path: "vminitd/Sources/LCShim"
+        ),
+        .target(
+            name: "Cgroup",
```
dcantah, line 281: I'm not super happy about this as I wanted to give the API we expose a better exploration, but if this is what we need to do for now that's okay with me.

dcantah (approved):

Package.swift
dkovba: Agreed, we can trim what doesn't need to cross the package boundary.

dcantah: I more meant if we could get away with not making Cgroup public as I haven't given a great deal of thought to the API outside of our immediate needs for it

dcantah: To be honest, I'd prefer if we had a ContainerizationCgroup target in the root of the repo if we are going to public-ify it.

dkovba: Moved the container `cgroup` setup from `vmexec` to `vminitd`.

dkovba: We've decided to revert https://github.com/apple/containerization/pull/742/commits/ad54d20907a819ca1a82809076df5b545f3d52f3. Marking as resolved.

dcantah (approved):

PR #747, opened 2026-05-21
Keep `CVersion` in the `vminitd` package

dkovba:
Moves `CVersion` back to the `vminitd` package. This is a follow-up for https://github.com/apple/containerization/pull/742.

### DIFF
--- a/vminitd/Sources/VminitdCore/AgentCommand.swift
+++ b/vminitd/Sources/VminitdCore/AgentCommand.swift
@@ -26,7 +26,6 @@ import GRPCCore
 import Logging
 import NIOCore
 import NIOPosix
-import Synchronization
 
 #if os(Linux)
 #if canImport(Musl)
@@ -45,8 +44,6 @@ public struct AgentCommand: AsyncParsableCommand {
 
     private static let foregroundEnvVar = "FOREGROUND"
     public static let vsockPort = 1024
-
-    public static let versionMetadata = Mutex<Logger.Metadata>([:])
 
     @OptionGroup var options: LogLevelOption
 
@@ -81,7 +78,7 @@ public struct AgentCommand: AsyncParsableCommand {
 
         signal(SIGPIPE, SIG_IGN)
 
-        log.info("vminitd booting", metadata: Self.versionMetadata.withLock { $0 })
+        log.info("vminitd booting", metadata: versionMetadata())
 
         // Set of mounts necessary to be mounted prior to taking any RPCs.
         // 1. /proc as the sysctl rpc wouldn't make sense if it wasn't there (NOTE: This is done before this method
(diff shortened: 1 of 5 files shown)


vminitd/Sources/vminitd/Application.swift
```diff
@@ -35,6 +36,8 @@ struct Application: AsyncParsableCommand {
     )
 
     static func main() async throws {
+        AgentCommand.versionMetadata.withLock { $0 = Self.versionMetadata() }
```
jglogan: Is this necessary? See comment for AgentCommand.swift.

vminitd/Sources/VminitdCore/AgentCommand.swift
```diff
@@ -79,14 +81,7 @@ public struct AgentCommand: AsyncParsableCommand {
...
 
-        let gitCommit = String(cString: CZ_get_git_commit())
-        let gitTag = String(cString: CZ_get_git_tag())
-        let buildTime = String(cString: CZ_get_build_time())
-        var metadata: Logger.Metadata = ["commit": "\(gitCommit)", "built": "\(buildTime)"]
-        if !gitTag.isEmpty {
-            metadata["tag"] = "\(gitTag)"
-        }
-        log.info("vminitd booting", metadata: metadata)
+        log.info("vminitd booting", metadata: Self.versionMetadata.withLock { $0 })
```
jglogan:
```suggestion
        log.info("vminitd booting", metadata: Application.versionMetadata.withLock { $0 })
```

dkovba: `VminitdCore` can't reference `Application` - the executable depends on the library, not the other way around.

vminitd/Sources/vminitd/Application.swift
dkovba: See a new commit as an alternative way of dropping the version metadata `Mutex` from `AgentCommand`.

vminitd/Sources/VminitdCore/AgentCommand.swift
jglogan: Why is AgentCommand in VminitCore? It makes more sense for all the swift-argument-parser commands to be in the same target as Application, whereas the core implements just the agent business logic.

vminitd/Sources/vminitd/Application.swift
jglogan: Putting it in logging metadata is going to add version to every log entry? That seems noisy.

dkovba: That was a bug. Fixed!

vminitd/Sources/VminitdCore/AgentCommand.swift
dkovba: The motivation for moving `VminitdCore` to the top-level package was specifically so that custom `vminitd` executables can reuse it. The CLI surface, including `AgentCommand`, `InitCommand`, `PauseCommand`, is part of what they want to reuse.

jglogan (approved):

PR #752, opened 2026-05-22
add IPv6 support

yibozhuang:
Extends the network plumbing to support per-interface IPv6 address configuration.
The `Interface` protocol supports `ipv6Address` and `ipv6Gateway`.
The agent's networking RPCs carry per-family fields via new `InterfaceAddress`, `LinkRoute`, and `DefaultRoute` types in `ContainerizationExtras`.
`NetlinkSession` adds IPv6 methods for address and route operations.


Sources/Containerization/VirtualMachineAgent+Interface.swift
dcantah, line 1: VirtualMachineAgent+Interface.swift?

Sources/Containerization/VirtualMachineAgent+Interface.swift
```diff
@@ -0,0 +1,84 @@
...
+                    ipv6Destination: needsIPv6LinkRoute ? ipv6Gateway : nil,
+                    ipv6Source: needsIPv6LinkRoute ? ipv6Address?.address : nil
+                )
+            )
+        }
+
+        if ipv4Gateway == nil && ipv6Gateway == nil {
+            logger?.debug("no gateway for \(name)")
+        }
+        try await routeAddDefault(
```
dcantah, line 79: Is this something we can skip if v4 and v6 gateway are both nil?

Sources/Containerization/VmnetNetwork.swift
```diff
@@ -137,11 +143,27 @@ public struct VmnetNetwork: Network {
...
+            ipv4Gateway: self.ipv4Gateway
+        )
+    }
+
+    /// Returns a new interface for use with a container, with an optional IPv6 address.
+    /// - Parameters:
+    ///   - id: The container ID.
+    ///   - ipv6Address: The IPv6 CIDR address to assign to the interface.
+    ///   - ipv6Gateway: The IPv6 gateway address for the default route.
+    public mutating func createInterface(_ id: String, ipv6Address: CIDRv6, ipv6Gateway: IPv6Address? = nil) throws -> Containerization.Interface? {
```
dcantah: Dumb question, this call was supposed to be the "just give me a NIC" call that `ContainerManager` can use. Can our allocator not be extended to also give us a v6 addr?

Sources/Containerization/VirtualMachineAgent+Interface.swift
yibozhuang: No we need to keep this call because #628 added support for installing a default route without a gateway. So skipping `routeAddDefault` when both v4 and v6 gateway are nil would cause a regression. The server-side handler at [Server+GRPC.swift:1290](https://github.com/apple/containerization/pull/752/changes#diff-bf2e96cd1d131c87b9a7d8235da85f591ca2cb977e4e272370d158c5638cfe69R1290) ensures the behavior is preserved.

Sources/Containerization/VmnetNetwork.swift
yibozhuang: Yes I think we can do the allocator + static v6 pattern here as well. The only thing is we also need to add `sysctl` to disable v6 auto-configuration (SLAAC) when using static v6 address. Working on this.

Sources/Containerization/VirtualMachineAgent+Interface.swift
yibozhuang: Fixed.

Sources/Containerization/VmnetNetwork.swift
yibozhuang: Fixed, added v6 allocator and added specifying v6 prefix to `VmnetNetwork` so now when calling `createInterface` will get both v4 and v6 address allocated.

crosbymichael (approved):

PR #762, opened 2026-06-02
Make LinuxPod.stopContainer return ExitStatus explicitly

noah-thor:
Update `LinuxPod.stopContainer` API to return the `ExitStatus` explicitly when the function call performs the kill and reaping itself. This lets callers observe the correct exit code without having to rely on other mechanisms.

### DIFF
--- a/Sources/Containerization/LinuxPod.swift
+++ b/Sources/Containerization/LinuxPod.swift
@@ -860,7 +860,14 @@ extension LinuxPod {
     }
 
     /// Stop a container from executing.
-    public func stopContainer(_ containerID: String) async throws {
+    ///
+    /// Returns the `ExitStatus` read from the container's process when this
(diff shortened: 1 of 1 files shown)


Sources/Containerization/LinuxPod.swift
```diff
@@ -860,7 +860,14 @@ extension LinuxPod {
     }
 
     /// Stop a container from executing.
-    public func stopContainer(_ containerID: String) async throws {
+    ///
```
dcantah, line 863: Sorry I know we talked about this, but does this offer anything that just calling `waitContainer` right after `stopContainer` returns not give? If anything I'd love if we had some functionality to cache the exit code so we don't actually need to do an RPC, and then `waitContainer` (and `wait` on `LinuxProcess` which it just trampolines to) just returns a cached integer

PR #770, opened 2026-06-15
docs: fix `traversal` in a public method's documentation

MojtabaHs:
Seems like a typo.
The change only affected a comment and does not disturb the code at all (backward compatible).


jglogan:
@MojtabaHs Good catch, thank you!

yibozhuang (approved):

PR #781, opened 2026-06-26
Relax escaping requirement for closures in Async{Lock|Mutex}

jamieQ:
I was looking into why https://github.com/apple/containerization/pull/696 was implemented, which was a fix for new warnings in the 6.4 compiler, and was confused why the diagnostic would be firing in the case it appeared to address. It turns out this was due to the `AsyncMutex.withLock` method taking an escaping closure parameter. However, the closure does not actually escape, so we can undo that prior workaround and make the API slightly more general by dropping the `@escaping` attribute. The same reasoning applies to `AsyncLock` as well, so for consistency I applied the change there too.

This does alter a public API though and I'm not sure what the source compatibility goals of this project are exactly, so let me know if you think that will be problematic.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -922,7 +922,7 @@ extension LinuxContainer {
                 agent: agent,
                 vm: startedState.vm,
                 logger: self.logger,
-                onDelete: { [weak self = self] in
+                onDelete: { [weak self] in
                     await self?.removeProcess(id: id)
                 }
             )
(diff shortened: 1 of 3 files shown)


crosbymichael:
@jamieQ Change looks good and makes a lot of sense to me. We do require signed commits. They are easy to setup with SSH keys if you get a chance.

jamieQ:
Thanks @crosbymichael – I tried to get the signed commit stuff configured correctly, but I think it may have invalidated the review. LMK if there's anything else I need to do.

PR #782, opened 2026-06-26
Add cloud-hypervisor VMM backend for Linux hosts

crosbymichael:
apple/containerization currently runs containers in per-container VMs on macOS hosts via Virtualization.framework. This adds a second VMM backend so the same Swift orchestration layer (LinuxContainer / LinuxPod / Vminitd gRPC contract) runs on Linux hosts via cloud-hypervisor + KVM.

**CloudHypervisor Swift package** (`Sources/CloudHypervisor/`) — a thin client for cloud-hypervisor's REST-over-UDS API, layered on AsyncHTTPClient. Endpoints cover VMM / VM lifecycle / hotplug (disk, fs, net, vsock, remove-device). Cross-platform (compiles on macOS for unit tests; consumed at runtime only by the Linux side of Containerization).

**CH backend in Containerization** — one cloud-hypervisor subprocess per VM, gated behind `#if os(Linux)`. CHVirtualMachineManager / CHVirtualMachineInstance mirror the VZ shape behind the existing VirtualMachineManager / VirtualMachineInstance protocol. CHProcess and VirtiofsdProcess manage the binaries; CHHotplugProvider handles virtio-blk and virtio-fs runtime hotplug (with one virtiofsd per unique source-hash tag, refcounted across containers).

**Linux host networking** — BridgeManager brings up a Linux bridge with an IPv4 subnet and (opt-in via `--enable-nat`) iptables MASQUERADE + scoped FORWARD rules. LinuxBridgedNetwork enslaves a fresh TAP per container to the bridge. State is recorded under `/run/containerization` so `cctl bridge delete` reverses exactly what create did. Bridge teardown verifies the link kind via sysfs to refuse deleting non-bridge interfaces.

**cctl run / bridge** — end-to-end Linux container run path (image pull, ext4 rootfs assembly, VM boot, container exec) plus `cctl bridge create|delete` for the host network plumbing.

**Build & dist** — `make linux-build` / `make linux-integration` build and exercise the host side inside an apple/container `--virtualization` dev container. `make dist-x86_64` produces a deployment tarball (cctl + cloud-hypervisor + virtiofsd + initfs + kernel) cross-compiled from the aarch64 dev container; pipeline documented in `docs/x86_64-build.md`. Static-musl C deps and the Zig cross compiler are pinned by SHA256.

The host orchestrator runs as root. Per-VM runtime state lives under `/run/containerization/ch/<UUID>` with mode 0700; UDS sockets inside are bound with mode 0600. Vminitd's gRPC channel inherits that trust boundary — socket-file perms are the auth.

Sandbox flags are upstream-secure by default. Two per-component opt-outs exist for the apple/container dev-container case (where the host seccomp profile SIGSYS-kills CH and virtiofsd):
- `CONTAINERIZATION_NO_CH_SECCOMP=1` — `cloud-hypervisor --seccomp false`.
- `CONTAINERIZATION_NO_VIRTIOFSD_SANDBOX=1` — `virtiofsd --sandbox none`. Each emits a one-shot `logger.warning` at process start. Legacy alias `CONTAINERIZATION_RELAXED_SANDBOX=1` flips both. cctl spawns both binaries with `setsid` and a minimal env allowlist (PATH / HOME / RUST_LOG / RUST_BACKTRACE) so the parent's secrets don't leak to children.

`make linux-integration` runs the cross-platform integration suite against a real cloud-hypervisor VM inside the dev container. Linux runs the cross-platform subset (`process true`/`false`/`echo hi`, virtiofs round-trip, hotplug); the macOS suite is unchanged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,91 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Build / Test / Format
+
+The project is built via `make`, not directly with `swift build`. Two Swift packages live in this repo: the root package (Containerization libraries + `cctl` + macOS-only integration binary) and `vminitd/` (the Linux guest init system, cross-compiled with the Static Linux SDK).
+
+- `make all` — build everything (`containerization` + `vminitd` + `init.ext4` rootfs in `bin/`). Default `BUILD_CONFIGURATION=debug`; pass `release` (or use `make release`) for optimized builds.
+- `make containerization` — build just the host-side Swift package (skips vminitd).
+- `make vminitd` — build vminitd / vmexec only. By default uses `LIBC=musl` via the Static Linux SDK; `make linux-build LIBC=glibc` builds via a Linux dev container.
+- `make test` — unit tests with code coverage. `make coverage` regenerates the coverage report.
+- `make integration` — runs `bin/containerization-integration`. Requires an in-repo kernel under `bin/` (`bin/vmlinux-arm64` on arm64, `bin/vmlinuz-x86_64` or `bin/vmlinux-x86_64` on x86_64); if absent, run `make fetch-default-kernel` to download the Kata-provided kernel for the host arch.
+- Single test: `swift test --filter ContainerizationOCITests.ReferenceTests/testParsing` (Swift Testing / XCTest filter syntax). Targets are listed in `Package.swift`.
+- `make linux-test` — runs `swift test` inside the Linux dev container (requires the `container` CLI from apple/container).
+- `make linux-build` — builds the host-side Swift package (incl. `cctl`, `Containerization`, and `CloudHypervisor`) inside the same Linux dev container. Use this to validate Linux portability of host-side code; the resulting `cctl` is what the cloud-hypervisor backend ships behind.
+- `make linux-integration` — runs the cross-platform integration suite against a real cloud-hypervisor VM inside the dev container (nested virt via apple/container's `--virtualization`). Requires a KVM-capable kernel at `kernel/vmlinux-arm64` (or `kernel/vmlinuz-x86_64` on x86_64 hosts) — build via `make -C kernel`; the kata-fetched kernel doesn't include KVM. Also requires `make fetch-cloud-hypervisor` and `make linux-build` to have been run first. Linux runs only the cross-platform subset (`process true`/`false`/`echo hi`); the macOS suite is unchanged.
+- `make fetch-cloud-hypervisor` — downloads the static `cloud-hypervisor` v52.0 (aarch64) binary into `bin/cloud-hypervisor` for the Linux integration tests.
+- `make build-cloud-hypervisor` / `make build-virtiofsd` — build patched `cloud-hypervisor` / `virtiofsd` from sources you have cloned into `.local/cloud-hypervisor` and `.local/virtiofsd` respectively. There is no fetch target — clone the upstream repos at the revision you want pinned. `build-virtiofsd` applies `scripts/patches/virtiofsd-skip-cap-drop-with-sandbox-none.patch` and is idempotent. Both run inside the same Linux dev container as `linux-integration` so the resulting binaries are aarch64-linux-gnu.
+- `make dist-x86_64` — assembles `bin/containerization-x86_64-<sha>.tar.gz` (cctl + cloud-hypervisor + virtiofsd + initfs.ext4 + kernel) for x86_64 Linux deployment, cross-compiled inside the aarch64 dev container via the Static Linux SDK (Swift) and `cargo zigbuild` (Rust). Prereqs: `.local/cloud-hypervisor` and `.local/virtiofsd` source checkouts (clone deliberately — no fetch target), and an x86_64 kernel built via `make -C kernel TARGET_ARCH=x86_64`. Per-stage rebuild env vars: `REBUILD_VMINITD=1`, `REBUILD_INITFS=1`, `REBUILD_CH=1`, `REBUILD_VIRTIOFSD=1`; cctl x86 always rebuilds. **Full pipeline, toolchain rationale, and troubleshooting in `docs/x86_64-build.md`.** The orchestrator is `scripts/build-dist-x86_64.sh`.
+- `make fmt` — applies `.swift-format` and refreshes license headers via hawkeye.
+- `make check` — formatting + license-header lint (this is what the pre-commit hook runs). Uses `.swift-format-nolint` for stricter linting.
+- `make pre-commit` — installs `scripts/pre-commit.fmt` as a git pre-commit hook.
+- `make protos` — regenerates `Sources/Containerization/SandboxContext/SandboxContext.{pb,grpc}.swift` from the `.proto`. Touch this whenever the proto changes; never hand-edit the generated files.
+- `make cross-prep` — installs Swiftly, the pinned Swift toolchain (see `.swift-version`), and the Static Linux SDK. Run once before the first build.
+
+`WARNINGS_AS_ERRORS=true` is the default for both packages. Don't disable it casually — CI builds with it on.
+
+## Architecture
+
+This is a **Swift library package** (not a CLI tool) that lets applications run Linux containers on Apple silicon by spawning a lightweight VM per container via `Virtualization.framework`. The corresponding end-user CLI lives in [`apple/container`](https://github.com/apple/container) and is **not** part of this repo. `cctl` here is a playground/example binary, not the shipping product.
+
+### The host ↔ guest split
+
+Every Linux container runs inside its own VM. The boundary between host (macOS) and guest (Linux) is the central architectural fact:
+
+- **Host side** (`Sources/`, `macOS` platform): orchestrates VMs through `Virtualization.framework` (`VZVirtualMachineInstance.swift`, `VZVirtualMachine+Helpers.swift`). The user-facing entry points are `LinuxContainer` (one container per VM) and `LinuxPod` (multiple containers in one VM, experimental). These build a `VMConfiguration`, boot the VM with the chosen `Kernel` and a rootfs containing `vminitd`, then drive the guest via gRPC.
+- **Guest side** (`vminitd/`, Linux platform): `vminitd` is PID 1 inside the VM. It exposes a gRPC service over **vsock** (default port `1024`) defined by `Sources/Containerization/SandboxContext/SandboxContext.proto`. `VminitdCore` implements that service: launching container processes, handling stdio over vsock, signal/event delivery, cgroups, mounts, and process lifecycle. By default it launches workloads via `vmexec` (a small helper that runs a single process inside the guest namespace); `runc` is used only when an OCI runtime path is supplied.
+
+The proto is the contract between the two halves. **The `.pb.swift` and `.grpc.swift` files in `SandboxContext/` are generated** — regenerate via `make protos` after changing `SandboxContext.proto`. Both host and guest depend on the same generated Swift via the path-dependency wiring in `vminitd/Package.swift` (`containerization` is a sibling path package).
+
+### VMM backends
+
+`Containerization` abstracts the VMM behind `VirtualMachineManager` / `VirtualMachineInstance`. Two backends ship in this repo, both inside the same `Containerization` target but gated by `#if`:
+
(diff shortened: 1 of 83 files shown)


Sources/cctl/RunCommand.swift
```diff
@@ -195,3 +195,338 @@ extension Application {
...
+        var cwd: String = "/"
+
+        @Argument(parsing: .captureForPassthrough)
+        var arguments: [String] = ["/bin/sh"]
+
+        func run() async throws {
+            #if arch(arm64)
+            let kernelPlatform = SystemPlatform.linuxArm
+            #else
+            let kernelPlatform = SystemPlatform.linuxAmd
```
jglogan, line 314: I think arch can be `x86_64`, `arm64`, `i386`, `arm`. It would be safer to explicitly test `x86_64` and error out in the `#else` branch.

Sources/cctl/RunCommand.swift
```diff
@@ -195,3 +195,338 @@ extension Application {
...
+
+            let containersRoot = Application.appRoot
+                .appendingPathComponent("containers")
+                .appendingPathComponent(id)
+            try FileManager.default.createDirectory(at: containersRoot, withIntermediateDirectories: true)
+            let rootfsPath = containersRoot.appendingPathComponent("rootfs.ext4")
+
+            var rootfsMount: Containerization.Mount
+            do {
+                let unpacker = EXT4Unpacker(blockSizeInBytes: fsSizeInMB.mib())
```
jglogan, line 356: Not the fault if this PR, and we can do this in a follow-on as it'd be a breaking change I think, but it looks like this parameter is misnamed.

This gets routed to `minDiskSize` in EXT4Formatter. We should call this `minDiskSizeInBytes`, `capacityBytes`, or something similar.

Sources/cctl/RunCommand.swift
```diff
@@ -195,3 +195,338 @@ extension Application {
...
+        var imageReference: String = "docker.io/library/alpine:3.16"
+
+        @Option(name: .long, help: "id for the container")
+        var id: String = "cctl"
+
+        @Option(name: [.customLong("cpus"), .customShort("c")], help: "Number of CPUs to allocate")
+        var cpus: Int = 2
+
+        @Option(name: [.customLong("memory"), .customShort("m")], help: "Amount of memory in MiB")
+        var memory: UInt64 = 1024
```
jglogan, line 224: rename to memoryInMB for clarity

Sources/cctl/RunCommand.swift
```diff
@@ -195,3 +195,338 @@ extension Application {
...
+                    let resolved =
+                        nameservers.isEmpty
+                        ? Self.readHostNameservers()
+                        : nameservers
+                    dnsConfig = DNS(nameservers: resolved)
+                }
+            }
+
+            let cpusCount = cpus
+            let memoryBytes = memory * 1024 * 1024
```
jglogan: `memoryInMB.mib()`

Sources/CloudHypervisor/CloudHypervisor+Error.swift
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import NIOHTTP1
+
+extension CloudHypervisor {
```
jglogan, line 20: Perhaps name file `CloudHypervisor+Error.swift`?

Sources/CloudHypervisor/CloudHypervisor+Client.swift
```diff
@@ -0,0 +1,169 @@
...
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Logging
+import NIOCore
+import NIOHTTP1
+import NIOPosix
+
+extension CloudHypervisor {
```
jglogan, line 23: `CloudHypervisor+Client.swift`?

Sources/CloudHypervisor/HTTPOverUDS.swift
```diff
@@ -0,0 +1,202 @@
...
+    /// Send an HTTP request and return the response.
+    ///
+    /// Translates AHC errors → ``CloudHypervisor/Error/transport(_:)`` so
+    /// callers see a uniform error type regardless of failure mode.
+    func send(
+        method: HTTPMethod,
+        uri: String,
+        body: Data?,
+        headers: HTTPHeaders = [:]
+    ) async throws -> HTTPResponse {
```
jglogan, line 101: Just a note, should we ever use swift-openapi-generator to create the client from the CH openapi spec, this stuff can move into swift-openapi-runtime middleware.

Sources/Containerization/BridgeManager.swift
```diff
@@ -0,0 +1,359 @@
...
+        do {
+            try session.linkDel(name: name)
+            log.info("deleted bridge \(name)")
+        } catch {
+            // ENODEV-like: nothing to do.
+            log.debug("linkDel \(name) returned \(error) (likely already absent)")
+        }
+
+        // 3. Restore ip_forward only if create()-with-NAT set it from 0.
+        if state?.natEnabled == true, state?.prevIpForward == "0" {
```
jglogan: Does this properly track all cases where multiple bridge networks exist with varying nat-enabled states? Is it possible to inadvertently set `ip_forward=0` even if there still exist nat-enabled bridges?

CLAUDE.md
```diff
@@ -0,0 +1,91 @@
+# CLAUDE.md
```
jglogan, line 1: `CLAUDE.md` or `AGENTS.md`? I don't see anything Claude-specific.

CLAUDE.md
```diff
@@ -0,0 +1,91 @@
...
+
+### The host ↔ guest split
+
+Every Linux container runs inside its own VM. The boundary between host (macOS) and guest (Linux) is the central architectural fact:
+
+- **Host side** (`Sources/`, `macOS` platform): orchestrates VMs through `Virtualization.framework` (`VZVirtualMachineInstance.swift`, `VZVirtualMachine+Helpers.swift`). The user-facing entry points are `LinuxContainer` (one container per VM) and `LinuxPod` (multiple containers in one VM, experimental). These build a `VMConfiguration`, boot the VM with the chosen `Kernel` and a rootfs containing `vminitd`, then drive the guest via gRPC.
+- **Guest side** (`vminitd/`, Linux platform): `vminitd` is PID 1 inside the VM. It exposes a gRPC service over **vsock** (default port `1024`) defined by `Sources/Containerization/SandboxContext/SandboxContext.proto`. `VminitdCore` implements that service: managing containers via runc, handling stdio over vsock, signal/event delivery, cgroups, mounts, and process lifecycle. `vmexec` is a small helper used to launch container processes from inside the guest.
```
jglogan: `runc`?

Sources/Containerization/LinuxProcess.swift
```diff
@@ -125,7 +125,7 @@ public final class LinuxProcess: Sendable {
 
 extension LinuxProcess {
     func setupIO(listeners: [VsockListener?]) async throws -> [FileHandle?] {
-        let handles = try await Timeout.run(seconds: 3) {
+        let handles = try await Timeout.run(seconds: 30) {
```
jglogan: Was this meant to be a temporary change? What prompted the 10x increase?

jglogan (commented):
62 files reviewed, 21 to go...

Sources/cctl/RunCommand.swift
crosbymichael: This matches the macOS side. I would opt to keep it simple as the declaration and calls are so close together.

Sources/cctl/RunCommand.swift
crosbymichael: Agreed, we can address this in a future PR.

Sources/Containerization/LinuxProcess.swift
crosbymichael: We should gate this to x86 only. Running tests with nested virt are much much slower but the benefit of nested virt to test this end to end outweighs the time cost at the moment.

CLAUDE.md
crosbymichael: we still use vmexec today, runc is an option

Sources/Containerization/HostDefaultRoute.swift
```diff
@@ -0,0 +1,67 @@
...
+    private static let RTF_GATEWAY: UInt32 = 0x0002
+
+    /// Parse the contents of `/proc/net/route` and return the iface for the
+    /// default route (destination 0.0.0.0 with `RTF_GATEWAY`). When multiple
+    /// default routes exist, the one with the lowest metric wins.
+    static func parseEgress(procNetRoute contents: String) -> String? {
+        var best: (iface: String, metric: UInt64)?
+        for (i, line) in contents.split(separator: "\n", omittingEmptySubsequences: true).enumerated() {
+            if i == 0 { continue }  // header
+            let cols = line.split(separator: "\t", omittingEmptySubsequences: false)
```
jglogan, line 39: Could we use netlink to get this data (and ipv6 routes should we need them later) instead?

```suggestion
            let cols = line.split(separator: "\t", omittingEmptySubsequences: false)
                .map { $0.trimmingCharacters(in: .whitespaces) }
```

Sources/cctl/RunCommand.swift
jglogan: That's fair. I realized that it's the same on macOS cctl, and container later.

Sources/cctl/RunCommand.swift
jglogan: https://github.com/apple/containerization/issues/787

Sources/Containerization/IptablesRules.swift
```diff
@@ -0,0 +1,102 @@
...
+/// ship the `iptables` binary as a shim over nftables and it's universally
+/// available.
+enum IptablesRules {
+    /// Add a rule unless it already exists. `args` is the rule body
+    /// excluding the leading action (`-A`/`-C`/`-D`).
+    ///
+    /// Implementation: run `iptables -C <args>` first; if exit 0, the rule
+    /// exists already — return. Otherwise run `iptables -A <args>` and
+    /// throw on non-zero exit.
+    static func ensure(table: String? = nil, args: [String]) throws {
```
jglogan, line 33: Should we run `iptables` with `-w`?

If BridgeManager owns a single bridge, can concurrent requests from multiple managers (as well as anything else running on the host that's mucking with nftables) result in failures that leave table state inconsistent with the bridge configuration state?

Sources/Containerization/CHVirtualMachineInstance.swift
```diff
@@ -0,0 +1,753 @@
...
+                try? FileManager.default.removeItem(at: entry.path)
+            }
+            throw error
+        }
+        _stdioPool.withLock { $0 = pool }
+        logger?.debug("vsock stdio pool prebound \(pool.count) ports starting at \(Self.stdioPoolBase)")
+    }
+
+    public func hotplug(_ block: Mount, id: String) async throws -> AttachedFilesystem {
+        try await hotplug.hotplug(block, id: id)
```
jglogan, line 478: Should we prevent hotplug calls once once in `.stopping`?

Sources/Containerization/CHVirtualMachineInstance.swift
```diff
@@ -0,0 +1,753 @@
...
+
+    public func hotplugVirtioFS(_ mounts: [Mount], id: String) async throws {
+        try await hotplug.hotplugVirtioFS(mounts, id: id)
+    }
+
+    public func releaseVirtioFS(id: String) async throws {
+        try await hotplug.releaseVirtioFS(id: id)
+    }
+
+    public func registerMounts(id: String, rootfs: AttachedFilesystem, additionalMounts: [Mount]) throws {
```
jglogan, line 493: Same here (we'd need to switch to async for asyncLock)

Sources/Containerization/CHVirtualMachineInstance.swift
```diff
@@ -0,0 +1,753 @@
...
+        guard current == .running else {
+            throw ContainerizationError(
+                .invalidState,
+                message: "vm is not running (state=\(current))"
+            )
+        }
+    }
+
+    public func listen(_ port: UInt32) throws -> VsockListener {
+        // Consume from the pre-bound pool (see `_stdioPool` doc).
```
jglogan, line 409: What happens if this is called while things are being torn down?

jglogan (approved):
Lgtm!

PR #783, opened 2026-06-27
fix(oci): render arm64 platform description without redundant v8

### LINKED ISSUE #1542: [Bug]: Inconsistent platform-string normalization (`arm64` ↔ `arm64/v8`) within a single build
### I have done the following

## Reproduction

Dockerfile (`Dockerfile.shared`):

```dockerfile
FROM alpine:3.20 AS builder
RUN mkdir -p /install && \
    head -c 1048576 /dev/urandom > /install/payload && \
    echo "chaos-1506 builder stage complete" > /install/marker

FROM alpine:3.20 AS runtime
COPY --from=builder /install /usr/local
CMD ["cat", "/usr/local/marker"]
```

Script:

```sh
#!/usr/bin/env bash
set -u

for i in 1 2 3 4 5; do
  container image rm chaos-1506-a:latest chaos-1506-b:latest >/dev/null 2>&1 || true

  container build --no-cache -t chaos-1506-a:latest -f Dockerfile.shared . \
    > /tmp/chaos-1506-a-${i}.log 2>&1 &
  PID_A=$!
  container build --no-cache -t chaos-1506-b:latest -f Dockerfile.shared . \
    > /tmp/chaos-1506-b-${i}.log 2>&1 &
  PID_B=$!

  wait $PID_A; wait $PID_B

  grep "linux/arm64" /tmp/chaos-1506-a-${i}.log
  grep "linux/arm64" /tmp/chaos-1506-b-${i}.log
done
```

The full script and fixture live at
[Sample Compose Files/CHAOS-1506-repro/](https://github.com/full-chaos/container-compose/pull/175/changes#diff-0791c445208755b4e8ec2b078465acb1e2c8dda90961dab53cbed0939c5565be) in
[https://github.com/full-chaos/container-compose](https://github.com/full-chaos/container-compose/pull/175/changes#diff-0791c445208755b4e8ec2b078465acb1e2c8dda90961dab53cbed0939c5565be)
(`Dockerfile.shared`, `repro.sh`). The reproduction does not depend on
container-compose — that's just where the repro happens to live for
convenience.

## Observed signal

Every per-build log shows the platform identifier changing between
stages of the **same** build. Example from one of build A's logs:

```
#6 [linux/arm64 builder 1/2] RUN mkdir -p /install ...
#7 [linux/arm64/v8 runtime 1/2] COPY --from=builder /install /usr/local
```

Builder stage: `linux/arm64`. Runtime stage of the same build:
`linux/arm64/v8`. The script never passes `--arch` or `--platform` —
`container build` defaulted both ends to the host arch but normalized
them inconsistently.

Reproduced 5/5 fresh-build iterations, every per-build log
(`/tmp/chaos-1506-a-{1..5}.log` and `/tmp/chaos-1506-b-{1..5}.log` —
10 total): each contains exactly one `linux/arm64/v8` occurrence
alongside `linux/arm64` ones.

## Production failure (Linear CHAOS-1506)

The same drift manifested as a hard build failure in a heavier user
workload (Python pip-install multi-stage Dockerfile):

```
=> CACHED [linux/arm64 builder 4/5] RUN pip install --prefix=/install ... 0.0s
=> [linux/arm64/v8 builder 3/5] COPY . . 19.6s
=> ERROR [linux/arm64/v8 runtime 2/5] COPY --from=builder /install /usr/local 0.0s
=> [linux/arm64 builder 4/5] RUN pip install --prefix=/install ... 2.1s
```

Same arm64/arm64/v8 drift, plus the `COPY --from=builder` resolved
to a state that triggered ERROR before the builder stage of the same
build completed. The lightweight repro above shows the drift but does
not reliably fire the COPY failure — likely because the alpine builder
finishes too quickly to widen the race window against runtime's
`COPY --from=builder` lookup.

## Hypothesis

The runtime stage's `COPY --from=builder` resolves the stage by name
via a platform-keyed lookup. When the lookup uses `arm64/v8` but the
layer was registered under `arm64` (or vice-versa), one of two things
happens:

- The stage cannot be found at all → COPY error fires before the
  builder stage completes.
- The lookup hits a sibling build's in-progress content-addressed cache
  entry (same content hash, different platform key) and fails on
  partial extraction under contention.

We were unable to confirm the precise mechanism from the client side —
the platform normalization happens inside the buildkit-shim.

## Client-side workaround

Users hit by the COPY failure can serialize build dispatches at the
client by running them sequentially (or via a tool that bounds
parallelism to 1). This avoids two concurrent `BuildCommand` clients
dialing the same buildkit container, which sidesteps the contention
that turns the platform drift into a hard COPY failure. It does **not**
fix the underlying normalization inconsistency.

## Ask

1. Confirm platform-string normalization is intended to be consistent
   across stages of one build invocation.
2. Investigate whether `COPY --from=<stage>` resolution is platform-keyed
   in a way that's vulnerable to the drift.
3. Once fixed, the client workaround can be removed.

## Logs

[container-build-shell-repro-build-B-run1.log](https://github.com/user-attachments/files/27610057/container-build-shell-repro-build-B-run1.log)
[container-build-shell-repro-build-A-run1.log](https://github.com/user-attachments/files/27610058/container-build-shell-repro-build-A-run1.log)

### Problem description

`container build` emits inconsistent platform identifiers across stages
of one build — `linux/arm64` for some stages and `linux/arm64/v8` for
others — even when the client never passes `--arch`, `--platform`, or
`--os`. The drift is benign in isolation, but appears to be the trigger
for a downstream `COPY --from=builder` failure under load (see
"Production failure" below).

Reproduces deterministically (5/5 fresh-build iterations) with two
parallel `container build` invocations against the same Dockerfile.
No third-party tooling involved.

### Environment

```markdown
- container CLI: 0.12.3 (build: release)
- macOS: Darwin 26.4.1
- Hardware: Apple Silicon Apple M4 Max
- Builder VM: default (2 CPU / 2048 MB; `container builder status`)
```

### Code of Conduct

Halvanhelv:
## Summary

`Platform.description` renders the same arm64 platform two different ways
depending on how the value was constructed:

```swift
Platform(arch: "arm64", os: "linux", variant: nil).description  // "linux/arm64"
Platform(arch: "arm64", os: "linux", variant: "v8").description // "linux/arm64/v8"
```

These are the **same** platform — `==`, `hash(into:)`, and `Set` membership
already treat an arm64 `nil` variant as equivalent to `"v8"` — yet they
serialize differently, so a single platform drifts between `linux/arm64` and
`linux/arm64/v8` across stages of one build (apple/container#1542).

## Relation to #764

#764 fixed the `Hashable` side of this: it stopped `hash(into:)` from using
`description` and canonicalized arm64 `nil` → `v8` in the hash. That worked
around the inconsistent `description` but did not fix it — its own summary
names `description` (`linux/arm64` vs `linux/arm64/v8`) as the root cause.
This PR fixes that remaining gap at the source.

## Change

Omit the redundant `v8` variant for arm64 when rendering `description`, so
equal arm64 platforms always describe as `linux/arm64` — matching how Docker
and containerd display the platform. Other variants (`arm/v7`) and
architectures (`amd64`) are unaffected.

Only the rendered `description` changes. The stored `variant` and the
`Codable` encoding are untouched, so OCI content digests remain stable.

## Testing

Added `OCIPlatformTests` cases for description consistency (equal arm64
platforms describe identically; `arm64/v8` renders as `linux/arm64`;
`arm/v7` and `amd64` preserved). `swift test --filter ContainerizationOCITests`
passes (55 tests); `swift format lint --strict` clean.

Closes apple/container#1542 (normalization-consistency aspect).

### DIFF
--- a/Sources/ContainerizationOCI/Platform.swift
+++ b/Sources/ContainerizationOCI/Platform.swift
@@ -47,13 +47,25 @@ public struct Platform: Sendable, Equatable {
         return .init(arch: normalized.arch, os: "linux", variant: normalized.variant)
     }
 
-    /// The computed description, for example, `linux/arm64/v8`.
+    /// The computed description, for example, `linux/amd64` or `linux/arm/v7`.
+    ///
+    /// `arm64`'s only defined variant is `v8`, which `==` and `hash` already treat as
+    /// equivalent to a `nil` variant. The redundant `v8` is therefore omitted so that
+    /// two equal arm64 platforms (one with `variant == nil`, one with `"v8"`) describe
+    /// identically as `linux/arm64`, rather than drifting between `arm64` and
+    /// `arm64/v8`.
     public var description: String {
         let architecture = architecture
-        if let variant = variant {
+        if let variant, !Self.isRedundantVariant(variant, for: architecture) {
             return "\(os)/\(architecture)/\(variant)"
         }
         return "\(os)/\(architecture)"
+    }
+
+    /// Whether `variant` is the canonical default for `architecture` and can be omitted
+    /// from the rendered description without losing information.
+    private static func isRedundantVariant(_ variant: String, for architecture: String) -> Bool {
+        architecture == "arm64" && variant == "v8"
     }
 
     /// The CPU architecture, for example, `amd64` or `arm64`.
--- a/Tests/ContainerizationOCITests/OCIPlatformTests.swift
+++ b/Tests/ContainerizationOCITests/OCIPlatformTests.swift
@@ -115,4 +115,40 @@ struct OCIPlatformTests {
         #expect(!matcher(Platform(arch: "arm64", os: "windows", variant: "v8")), "matcher must reject a differing OS")
         #expect(matcher(Platform(arch: "arm64", os: "linux", variant: "v8")), "matcher must accept the same OS with an implied v8 variant")
     }
+
+    // MARK: - description consistency
+
+    @Test func arm64_nilAndV8_sameDescription() {
+        let withoutVariant = Platform(arch: "arm64", os: "linux", variant: nil)
+        let withV8 = Platform(arch: "arm64", os: "linux", variant: "v8")
+        #expect(
+            withoutVariant.description == withV8.description,
+            "equal arm64 platforms must produce the same description"
+        )
+    }
+
+    @Test func arm64_descriptionDropsRedundantV8() {
+        let withV8 = Platform(arch: "arm64", os: "linux", variant: "v8")
+        #expect(withV8.description == "linux/arm64", "arm64/v8 is canonical arm64, rendered without the redundant variant")
+    }
+
+    @Test func arm64_nilVariantDescription() {
+        let withoutVariant = Platform(arch: "arm64", os: "linux", variant: nil)
+        #expect(withoutVariant.description == "linux/arm64")
+    }
+
+    @Test func arm64_fromStringWithV8DescriptionIsCanonical() throws {
+        let parsed = try Platform(from: "linux/arm64/v8")
+        #expect(parsed.description == "linux/arm64", "parsing arm64/v8 then describing must yield the canonical short form")
+    }
+
+    @Test func arm_v7_descriptionKeepsVariant() {
+        let armv7 = Platform(arch: "arm", os: "linux", variant: "v7")
+        #expect(armv7.description == "linux/arm/v7", "non-redundant variants such as arm/v7 must be preserved")
+    }
+
+    @Test func amd64_descriptionUnaffected() {
+        let amd64 = Platform(arch: "amd64", os: "linux")
+        #expect(amd64.description == "linux/amd64")
+    }
 }


adityaramani:
Thanks for the change! Can you please rebase the change before we merge

Halvanhelv:
> Thanks for the change! Can you please rebase the change before we merge

Done

adityaramani (approved):
LGTM! Only nit I have is do we need to remove the references to the issues in the code. Will wait for another maintainer to weigh in

Halvanhelv:
> LGTM! Only nit I have is do we need to remove the references to the issues in the code. Will wait for another maintainer to weigh in

Great, I'll try to delete it quickly if that's the decision

jglogan:
@Halvanhelv Yep, go ahead and take the issue reference in the test comment and we'll build, approve, and merge. Thank you!

jglogan (changes requested):
Just the issue comment, otherwise looks good.

Halvanhelv:
> Just the issue comment, otherwise looks good.

done

jglogan (approved):

jglogan:
@Halvanhelv Thanks! Merged.

PR #793, opened 2026-07-07
perf: index EXT4 FileTree children by name to avoid O(n^2) unpack

kkilchrist:
## What

`FileTree.lookup` resolved each path component by linearly scanning the node's `children` array. This changes the node's child storage to an `OrderedDictionary<String, Ptr<FileTreeNode>>` (from swift-collections, which is already a package dependency) keyed by name, so `lookup` resolves each component in O(1) while iteration keeps the existing insertion order.

## Why

`EXT4.Formatter.create` calls `lookup` for every entry (and once per ancestor directory). With a linearly-scanned array, populating a directory with K entries is O(K²), which dominates unpack time for images with large directories (`node_modules`, `/usr/lib`, apt lists, …).

## Design

`childrenByName` is the single source of truth; `children` remains available as an ordered view (`.values`), so existing iterate/map call sites are unchanged. Mutation goes through two helpers on `FileTreeNode` — `addChild(_:)` and `removeChild(named:)` — used by every mutation site: `Formatter.link`, `Formatter._unlink`, `Formatter.create`, `FileTreeNode.init`, and the reader's tree build in `EXT4+Reader`. A single keyed structure also makes duplicate-name children unrepresentable.

*(An earlier revision of this PR added a separate `childIndex` dictionary mirroring the `children` array; per review feedback it now uses a single `OrderedDictionary` instead — same lookup performance, less memory, and no two structures to keep in sync.)*

## Results

Measured on Apple Silicon unpacking `louislam/uptime-kuma:1` (40,834 files / 438 MB):

- Optimized build: **~15s → ~9s (~40% faster)**.
- The relative improvement grows with directory size; in unoptimized/debug builds the absolute difference is far larger.

A single-directory microbenchmark (30,000 `create` calls into one directory, release build) runs in ~0.33s vs ~1.9s before this change, with peak RSS slightly lower than the array+index revision.

## Testing

`swift test --filter ContainerizationEXT4Tests` — all pass (62 tests), including the whiteout, hardlink, symlink, and replace-semantics cases that exercise the mutated sites.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -106,6 +106,7 @@ let package = Package(
             name: "ContainerizationEXT4",
             dependencies: [
                 "ContainerizationArchive",
+                .product(name: "OrderedCollections", package: "swift-collections"),
                 .product(name: "SystemPackage", package: "swift-system"),
                 "ContainerizationOS",
             ],
--- a/Sources/ContainerizationEXT4/EXT4+FileTree.swift
+++ b/Sources/ContainerizationEXT4/EXT4+FileTree.swift
@@ -15,6 +15,7 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import OrderedCollections
 import SystemPackage
 
 extension EXT4 {
@@ -22,7 +23,11 @@ extension EXT4 {
         class FileTreeNode {
             let inode: InodeNumber
             let name: String
-            var children: [Ptr<FileTreeNode>] = []
+            // Children keyed by name for O(1) lookup, preserving insertion order.
+            private(set) var childrenByName: OrderedDictionary<String, Ptr<FileTreeNode>> = [:]
+            var children: OrderedDictionary<String, Ptr<FileTreeNode>>.Values {
+                childrenByName.values
+            }
             var blocks: (start: UInt32, end: UInt32)?
             var additionalBlocks: [(start: UInt32, end: UInt32)]?
             var link: InodeNumber?
@@ -39,16 +44,17 @@ extension EXT4 {
             ) {
                 self.inode = inode
                 self.name = name
-                self.children = children
                 self.blocks = blocks
                 self.additionalBlocks = additionalBlocks
                 self.link = link
                 self.parent = parent
+                for child in children {
+                    self.addChild(child)
+                }
             }
 
             deinit {
-                self.children.removeAll()
-                self.children = []
+                self.childrenByName.removeAll()
                 self.blocks = nil
                 self.additionalBlocks = nil
                 self.link = nil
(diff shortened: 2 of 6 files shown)


jglogan:
@kkilchrist Did you consider replacing Array+Dictionary with OrderedDictionary from swift-collections? Would that improve performance while reducing additional memory consumption?

kkilchrist:
@jglogan 
I hadn't considered it. I kept the original change additive as an outside contributor. But you're right that OrderedDictionary is strictly better.

A small benchmark on my side shows OrderedDictionary has equivalent performance to and lower memory footprint than the Array + Index in my original proposal. (Array+Index / OrderedDictionary both ~0.33 s compared to ~1.9 s for upstream implementation). 

Since ext4 materialization is one of the key differences vs overlayfs layer-composition, improving performance here will narrow the latency gap on first-pull cold starts. 

PR updated with suggestion. Thanks for the feedback!

jglogan:
@kkilchrist Could you make sure your commits are signed and verified?

jglogan:
@kkilchrist Just dropping another reminder that we can't merge the change without signed, verified commits. Thanks!

kkilchrist:
Hi @jglogan, digging into this today. Sorry for the delay. I'll update again when the commits are signed and all tests pass. 

Edit: should be signed now. I think the failing tests are not related to these changes.

jglogan:
@kkilchrist thanks! I'm going to run the container integration suite with a `swift package edit` for containerization and your patch applied, as a final sanity check. I just ran the suite a few secs ago so it'll be interesting to see how much impact there is on wall clock time for the tests.

jglogan:
All the tests are running fine in container (no surprise, everything looks sound in the PR).

Running with a release build of cctl I see the following unpack times (sample size 1, I ran a some a few times with no significant difference). We don't see a return on optimization until we hit tens of thousands of files, but the slightly more expensive insert isn't hurting us for medium sized images either.

| Image | N (approx) | t (before, sec) | t (after, sec) |
|---|---|---|---|
|  `docker.io/library/alpine:3.24.1` | 500 | 0.044 | 0.045 |
|  `docker.io/library/swift:6.3.3` | 15,000 | 5.685 | 5.799 |
|  `docker.io/continuumio/anaconda3:2024.10-1` | 120,000 | 23.406 | 10.589 |

jglogan (approved):

jglogan:
@kkilchrist Thanks for the optimization, and thank you for updating this in time for the next release!

kkilchrist:
@jglogan very welcome! Thanks for the update on the profiling too.

PR #799, opened 2026-07-11
fix(copy): keep lifecycle usable after missing source

### LINKED ISSUE #1927: [Bug]: container cp on nonexistent source path hangs indefinitely; exec/stop on same container then hang too
### I have done the following

### Steps to reproduce

1. container run -it --name test ubuntu:latest bash
2. container cp test:/existing/file.txt ~/local/file.txt
   → works fine, returns fast
3. container cp test:/path/that/does/not/exist ~/local/foo.txt
   → hangs indefinitely, no error, no timeout
4. Ctrl-C to abort
5. container stop test
   → also hangs indefinitely
5. container system stop
   → also hangs indefinitely

Only recovery found: restart macos.

### Problem description

`container cp` on a source path that doesn't exist inside the container hangs
forever instead of returning an error immediately. Worse, once this happens every subsequent
command against that same container — exec -it, stop, system stop — also
hangs indefinitely.

Expected behavior:
- cp on a missing path should fail fast with a clear "no such file or
  directory" error.
- Even if a single operation does hang, it should not take down the whole
  container's control channel — stop should still work, or at minimum
  should time out with a clear error rather than hanging forever.

Actual behavior:
- cp hangs with no timeout.
- exec and stop on the same container then hang too.
- Only way to recover is restart macos.

### Environment

```markdown
- OS: macOS 26.5.2 (Tahoe)
- Hardware: Apple Silicon (M4)
- Container: Container CLI version 1.1.0 (installed via Homebrew)
- Image: docker.io/library/ubuntu:latest (Ubuntu 26.04 LTS, arm64)
```

### Code of Conduct

stephenlclarke:
## Summary

Finish copy-out coordination streams when guest-side source validation fails, preserve the original guest error across the task-group race, and release the container state lock so later lifecycle operations remain usable.

## Motivation

[apple/container#1927](https://github.com/apple/container/issues/1927) reports that `container cp` hangs indefinitely when the guest source path does not exist. The same container then cannot be used by `exec`, `stop`, `delete`, or system shutdown because `LinuxContainer.copyOut` still owns its state lock.

The producer task receives the guest error before emitting metadata or establishing a vsock connection. Its sibling task waits on those asynchronous streams and does not unblock merely because the task group is cancelled. The group therefore cannot unwind. Closing both streams fixes the deadlock, while a small synchronized error state ensures the consumer cannot replace the useful guest error with an internal `no metadata` or `vsock connection not established` error.

## Changes

- Finish the metadata continuation and vsock listener on every producer failure.
- Keep outer cleanup idempotent for successful and failed transfers.
- Preserve the original guest error if either consumer guard observes stream termination first.
- Add a VM integration regression that bounds missing-source copy, rejects coordination fallback errors, and runs a successful exec against the same container afterward.

## Validation

```sh
swift build --disable-automatic-resolution \
  --product containerization-integration \
  -Xswiftc -warnings-as-errors
swift test --enable-code-coverage \
  --disable-automatic-resolution \
  -Xswiftc -warnings-as-errors
make check
./bin/containerization-integration \
  --kernel ./bin/vmlinux-arm64 \
  --max-concurrency 4 \
  --filter 'container copy out missing source does not block lifecycle'
./bin/containerization-integration \
  --kernel ./bin/vmlinux-arm64 \
  --max-concurrency 4 \
  --filter 'container copy out'
```

- All 569 Swift tests in 79 suites pass with coverage instrumentation and warnings treated as errors.
- Swift formatting and Hawkeye license checks pass.
- The exact missing-source VM regression passes three consecutive runs, returns the original `notFound` guest-path error, and completes a subsequent exec on the same container.
- All 3 matching copy-out VM integration cases pass together.
- The committed diff has no whitespace errors.

## Compatibility

There is no public API or successful-transfer behavior change. Missing guest paths now fail promptly and leave the container usable. `apple/container` can surface the lower-runtime error without adding its own lock or stream cleanup.

## Related

- Fixes [apple/container#1927](https://github.com/apple/container/issues/1927).
- Root-cause analysis: [apple/container#1927 comment](https://github.com/apple/container/issues/1927#issuecomment-4925733977).

## Release Note Highlight

- Fixes [apple/container#1927](https://github.com/apple/container/issues/1927): copying a missing container path now fails promptly without blocking later exec, stop, or delete operations.

### DIFF
--- a/Sources/Containerization/LinuxContainer.swift
+++ b/Sources/Containerization/LinuxContainer.swift
@@ -26,6 +26,18 @@ import SystemPackage
 
 import struct ContainerizationOS.Terminal
 
+private final class CopyOutProducerError: Sendable {
+    private let value = Mutex<(any Error)?>(nil)
+
+    func store(_ error: any Error) {
+        value.withLock { $0 = error }
+    }
+
+    func load() -> (any Error)? {
+        value.withLock { $0 }
+    }
+}
+
 /// `LinuxContainer` is an easy to use type for launching and managing the
 /// full lifecycle of a Linux container ran inside of a virtual machine.
 public final class LinuxContainer: Container, Sendable {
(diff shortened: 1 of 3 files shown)


wooii:
I believe this PR fixed https://github.com/apple/container/issues/1927.

Built `container` from source with `containerization` pointed at this PR #799 's branch (commit `9c5b9ee`) via `swift package edit`.

**Env:** macOS 26.5.2, Apple Silicon (M4), Swift 6.3.3 / 6.3-RELEASE toolchains.

**Results:**
- `container cp` a nonexistent path → fails immediately with a clear error instead of hanging. ✅
    ```
   Error: failed to copy from container test (cause: "internalError: "failed to copy from container test (cause: "unknown: "notFound: "copy: path not found '/run/container/test/rootfs/...'""")"")
   ```
- Copying an existing path (e.g. `/etc/os-release`) → still works, returns promptly. ✅
- `container exec` on the same container right after the failed copy → works, no hang. ✅
- `container stop` on the same container → hit an XPC "Connection interrupted" error, but this looks like a separate issue. Haven't fully isolated whether it's related to this fix. ⚠️
  ```
  Error: internalError: "failed to stop container" (cause: "internalError: "failed to stop container test (cause: "interrupted: "XPC connection error: Connection interrupted"")"")
  Ensure container system service has been started with container system start.
  ```

Core issue from [#1927](https://github.com/apple/container/issues/1927) (indefinite hang on `container cp` to a missing path, and the resulting `exec`/`stop` lock-up) is resolved. Thanks @stephenlclarke and @lakshitsoni26!

PR #810, opened 2026-07-22
Build vminitd and initfs inside the dev container

crosbymichael:


Tests/ContainerizationOSTests/EpollTests.swift
```diff
@@ -30,6 +30,10 @@ import Glibc
 @Suite("Epoll tests")
 final class EpollTests {
 
+    // write() to a pipe whose read end is closed must not kill the
+    // `swift test` process with SIGPIPE on Linux. See ignoreSIGPIPEForTests().
+    init() { ignoreSIGPIPEForTests() }
```
noah-thor, line 35: nit: Are there any tests we don't want to ignore this? Could make this a test trait, but functionality will be the same, just maybe more "idiomatic"

Makefile
```diff
@@ -294,26 +294,60 @@ ifeq ($(UNAME_S),Darwin)
 	@codesign --force --sign - --timestamp=none --entitlements=signing/vz.entitlements bin/containerization-integration
 endif
 
+# Shell fragments run inside the Linux dev container (see linux_run). Kept as
+# variables so `vminitd` (compile only) and `init` (compile + build the initfs
+# in a single container run) don't duplicate the command.
+VMINITD_BUILD_CMD = make -C vminitd BUILD_CONFIGURATION=$(BUILD_CONFIGURATION) WARNINGS_AS_ERRORS=$(WARNINGS_AS_ERRORS)
+INITFS_BUILD_CMD = ./scripts/build-initfs.sh --vminitd vminitd/bin/vminitd --vmexec vminitd/bin/vmexec --ext4 bin/initfs.ext4 --tar bin/init.rootfs.tar.gz
```
noah-thor, line 301: Would it be beneficial to make this script a make target?

Sources/cctl/RootfsCommand.swift
```diff
@@ -34,147 +30,40 @@ extension Application {
         )
 
         struct Create: AsyncParsableCommand {
-            @Option(name: [.short, .customLong("add-file")], help: "Additional file to add (format src-path:dst-path)")
-            var addFiles: [String] = []
-
-            @Option(name: .customLong("ext4"), help: "The path to an ext4 image to create.")
-            var ext4File: String?
+            static let configuration = CommandConfiguration(
```
noah-thor, line 33: For my understanding, the refactor of this class is because now we don't need to do these operations from a darwin host and thus they can be done "natively" in linux during the `build-initfs`?

Tests/ContainerizationOSTests/EpollTests.swift
crosbymichael: Naw, we are safe to ignore SIGPIPE for our applications.

Sources/cctl/RootfsCommand.swift
crosbymichael: Ya, we can both compile vminitd and build the ext4 based initfs inside the container.

noah-thor (approved):

PR #814, opened 2026-07-24
Validate symlink targets in archiveReader and mask high mode bits

katiewasnothere:
This PR does two things: 
- Prevents target paths for symlinks from linking outside of the extraction root of the archive. Tar entries that attempt to link outside of the extraction root are added to the list of rejected paths returned from extractContents.
- Masks the high mode bits when setting file attributes for consistency with mode at file creation

### DIFF
--- a/Sources/ContainerizationArchive/ArchiveReader.swift
+++ b/Sources/ContainerizationArchive/ArchiveReader.swift
@@ -369,8 +369,17 @@ extension ArchiveReader {
                 guard let targetPath = (entry.symlinkTarget.map { FilePath($0) }) else {
                     return false
                 }
+                guard !targetPath.isAbsolute else {
+                    return false
+                }
                 var symlinkCreated = false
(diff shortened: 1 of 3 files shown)


ans736-gif (approved):

iabdullah215:
`lexicallyNormalized()` crashes in debug builds on some traversal targets (swift-system 1.7.5 assertion). Release is fine, so not an escape, just a debug crash on malicious symlinks.
Repro:
```swift
import SystemPackage
_ = FilePath("x/../../../../y").lexicallyNormalized()  // FilePathParsing.swift:231: Assertion failed
```
Consider checking escape by walking components with a depth counter (reject if it goes negative) instead of normalizing.

katiewasnothere:
Closing. We've decided we do not want to limit the symlink targets here. We already use O_NOFOLLOW in the archive code when opening files.

PR #827, opened 2026-07-30
Align CIDR containment and bounds with CIDRv4 and CIDRv6

### LINKED ISSUE #826: [Bug]: CIDR.contains excludes members of a block written from a host address
### I have done the following

- [x] I reproduced the issue against `main` at `ff44a5b683c80fceab875dba8a20ed24d7648c07`

### Steps to reproduce

`CIDR` keeps the address exactly as parsed, so a block written from a host address keeps its host bits set. Containment then compares that unmasked value against the masked probe.

Add this to `Tests/ContainerizationExtrasTests/TestCIDR.swift` and run `swift test --filter TestCIDR`:

```swift
@Test func reproduce() throws {
    let block = try CIDR("192.168.1.100/24")
    #expect(block.contains(try IPAddress("192.168.1.100")))  // fails
    #expect(block.contains(try IPAddress("192.168.1.0")))    // fails

    // the type it wraps gets this right
    let concrete = try CIDRv4("192.168.1.100/24")
    #expect(concrete.contains(try IPv4Address("192.168.1.100")))  // passes
}
```

### Current behavior

The wrapper disagrees with the type it wraps, in both families:

| expression | `CIDR` | `CIDRv4` / `CIDRv6` |
|---|---|---|
| `192.168.1.100/24` contains `192.168.1.100` | `false` | `true` |
| `192.168.1.100/24` contains `192.168.1.0` | `false` | `true` |
| `10.1.2.3/16` contains `10.1.99.99` | `false` | `true` |
| `2001:db8::1234/64` contains `2001:db8::1` | `false` | `true` |

`CIDR.contains` compares an unmasked stored address (`Sources/ContainerizationExtras/CIDR.swift:116` and `:118`):

```swift
return network.value == (ip.value & prefix.prefixMask32)
```

whereas `CIDRv4.contains` masks both sides:

```swift
(address.value & prefix.prefixMask32) == (ip.value & prefix.prefixMask32)
```

A block already written in network form (`192.168.1.0/24`) is unaffected, which is why the current tests pass: every containment case in `TestCIDR.swift` uses a network address.

There is a second defect in the same accessors. `lower` drops the IPv6 zone while `upper` keeps it (`CIDR.swift:94` vs `:105`, and `CIDRv6.swift:81` vs `:87`):

```
2001:db8::5%lo0/126  ->  lower = 2001:db8::4       (zone lost)
                         upper = 2001:db8::7%lo0
contains(lower) = false
```

Because `contains` compares zones, the missing zone on its own is enough to place the bound outside its own block.

### Expected behavior

`CIDR` should agree with `CIDRv4` and `CIDRv6` on both containment and bounds: mask the stored address before comparing, and carry the address's zone into `lower` the way `upper` already does.

### Environment

```markdown
macOS 26.6 (25G72), Xcode 26.6, Swift 6.3, arm64
main @ ff44a5b683c80fceab875dba8a20ed24d7648c07
```

### Relevant log output

```shell
✘ Expectation failed: block.contains(try IPAddress(ip))
    cidr → "192.168.1.100/24", ip → "192.168.1.100"
✘ Expectation failed: (wrapper.contains(.v4(ip)) → false) == (concrete.contains(ip) → true)
    cidr → "192.168.1.100/24"
✘ Expectation failed: (block → 2001:db8::5%lo0/126).contains(block.lower → 2001:db8::4)
```

### Code of Conduct

vyncint:
Fixes #826.

## What

`CIDR` keeps the address exactly as parsed, so a block written from a host address keeps its host bits set. `contains` compared that unmasked stored value against the masked probe, so the wrapper reported that a block excluded its own members while the types it wraps answered the same query correctly.

| expression | before | `CIDRv4` / `CIDRv6` |
|---|---|---|
| `192.168.1.100/24` contains `192.168.1.100` | `false` | `true` |
| `192.168.1.100/24` contains `192.168.1.0` | `false` | `true` |
| `10.1.2.3/16` contains `10.1.99.99` | `false` | `true` |
| `2001:db8::1234/64` contains `2001:db8::1` | `false` | `true` |

Masking both sides matches what `CIDRv4.contains` and `CIDRv6.contains` already do. A block written in network form was unaffected, which is why this went unnoticed: every containment case in `TestCIDR.swift` used a network address.

The same accessors carried a second defect. `lower` dropped the IPv6 zone while `upper` kept it, and because `contains` compares zones, a zoned block excluded its own lower bound:

```
2001:db8::5%lo0/126  ->  lower = 2001:db8::4       (zone lost)
                         upper = 2001:db8::7%lo0
```

`lower` now carries the address's zone the way `upper` already did, in both copies of the logic: the `CIDR` wrapper and `CIDRv6`.

One caller needed a matching change. `VmnetNetwork.configurePrefixV6` renders `lower` into `inet_pton`, which rejects a zone suffix, so it was relying on `lower` stripping it. Per review, the vmnet API now leaves that to the caller: the helper renders `lower` as-is and checks the `inet_pton` result, rejecting a prefix it cannot parse rather than quietly accepting a zoned one. That check also closes a hole older than this PR — the result was discarded, so any address `inet_pton` refused left the buffer zeroed and configured `::`. Every in-repo caller already passes a zoneless prefix, so no call site changes.

## Verification

- 9 new tests in `TestCIDR.swift`: containment across v4 and v6 host-address blocks, non-members still excluded, the wrapper agreeing with `CIDRv4`/`CIDRv6` over a probe set, zoned bounds contained in their own block, zone rendering, and unzoned bounds still carrying no zone.
- Negative control: reverting only `CIDR.swift` and `CIDRv6.swift` while keeping the tests fails **7 of the 9**, with the wrong values visible in the output, for example `(block → 2001:db8::5%lo0/126).contains(block.lower → 2001:db8::4)` and `(wrapper.contains(.v4(ip)) → false) == (concrete.contains(ip) → true)`. The 2 that still pass are the controls written to pass either way: non-members stay excluded, and unzoned bounds are untouched.
- Full suite passes: 591 tests in 81 suites.
- `make containerization` builds clean with the default `WARNINGS_AS_ERRORS=true`.
- `swift format lint --strict --recursive --configuration .swift-format-nolint Sources Tests` exits 0, `swift format` leaves all four files unchanged, and `make check-licenses` reports no missing headers.
- The vmnet helper itself has no unit coverage: it is `#if os(macOS)`, private, and reachable only through `VmnetNetwork.init`, which creates a real vmnet network. It is exercised by the integration suite, which needs the vmnet entitlement.

The public shape of `CIDR`, `CIDRv4` and `CIDRv6` is untouched; only the values these accessors compute change.

### DIFF
--- a/Sources/Containerization/VmnetNetwork.swift
+++ b/Sources/Containerization/VmnetNetwork.swift
@@ -291,7 +291,10 @@ public struct VmnetNetwork: Network {
 
     private static func configurePrefixV6(_ config: vmnet_network_configuration_ref, prefixV6: CIDRv6) throws {
         var p = in6_addr()
-        inet_pton(AF_INET6, prefixV6.lower.description, &p)
+        // `inet_pton` rejects a zone suffix, so render the network address on
+        // its own. A vmnet prefix is never link-scoped, so dropping the zone
+        // here loses nothing.
+        inet_pton(AF_INET6, IPv6Address(prefixV6.lower.value).description, &p)
 
         guard vmnet_network_configuration_set_ipv6_prefix(config, &p, prefixV6.prefix.length) == .VMNET_SUCCESS else {
             throw ContainerizationError(.internalError, message: "failed to set IPv6 prefix \(prefixV6) for network")
--- a/Sources/Containerization/VmnetNetwork.swift
+++ b/Sources/Containerization/VmnetNetwork.swift
@@ -291,10 +291,13 @@ public struct VmnetNetwork: Network {
 
     private static func configurePrefixV6(_ config: vmnet_network_configuration_ref, prefixV6: CIDRv6) throws {
         var p = in6_addr()
-        // `inet_pton` rejects a zone suffix, so render the network address on
-        // its own. A vmnet prefix is never link-scoped, so dropping the zone
-        // here loses nothing.
-        inet_pton(AF_INET6, IPv6Address(prefixV6.lower.value).description, &p)
+        // `inet_pton` rejects a zone suffix, and a vmnet prefix is never
+        // link-scoped, so the caller is responsible for supplying a zoneless
+        // prefix. Check the result: ignoring it leaves `p` zeroed and
+        // configures `::` on anything that fails to parse.
+        guard inet_pton(AF_INET6, prefixV6.lower.description, &p) == 1 else {
+            throw ContainerizationError(.invalidArgument, message: "invalid IPv6 prefix \(prefixV6) for network")
+        }
 
         guard vmnet_network_configuration_set_ipv6_prefix(config, &p, prefixV6.prefix.length) == .VMNET_SUCCESS else {
             throw ContainerizationError(.internalError, message: "failed to set IPv6 prefix \(prefixV6) for network")
--- a/Sources/ContainerizationExtras/CIDR.swift
+++ b/Sources/ContainerizationExtras/CIDR.swift
@@ -91,7 +91,7 @@ public enum CIDR: CustomStringConvertible, Equatable, Sendable, Hashable {
         case (.v4(let addr, let prefix)):
             return .v4(IPv4Address(addr.value & prefix.prefixMask32))
         case (.v6(let addr, let prefix)):
-            return .v6(IPv6Address(addr.value & prefix.prefixMask128))
+            return .v6(IPv6Address(addr.value & prefix.prefixMask128, zone: addr.zone))
         }
     }
 
@@ -113,9 +113,9 @@ public enum CIDR: CustomStringConvertible, Equatable, Sendable, Hashable {
     public func contains(_ ip: IPAddress) -> Bool {
         switch (self, ip) {
         case (.v4(let network, let prefix), .v4(let ip)):
-            return network.value == (ip.value & prefix.prefixMask32)
+            return (network.value & prefix.prefixMask32) == (ip.value & prefix.prefixMask32)
         case (.v6(let network, let prefix), .v6(let ip)):
-            return (network.zone == ip.zone) && (network.value == (ip.value & prefix.prefixMask128))
+            return (network.zone == ip.zone) && ((network.value & prefix.prefixMask128) == (ip.value & prefix.prefixMask128))
         default:
             return false
         }
--- a/Sources/ContainerizationExtras/CIDRv6.swift
+++ b/Sources/ContainerizationExtras/CIDRv6.swift
@@ -78,7 +78,7 @@ public struct CIDRv6: CustomStringConvertible, Equatable, Sendable, Hashable {
     /// The lowest address in this CIDR block
     @inlinable
     public var lower: IPv6Address {
-        IPv6Address(address.value & prefix.prefixMask128)
+        IPv6Address(address.value & prefix.prefixMask128, zone: address.zone)
     }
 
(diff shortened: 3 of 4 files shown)


Sources/Containerization/VmnetNetwork.swift
```diff
@@ -291,7 +291,10 @@ public struct VmnetNetwork: Network {
 
     private static func configurePrefixV6(_ config: vmnet_network_configuration_ref, prefixV6: CIDRv6) throws {
         var p = in6_addr()
-        inet_pton(AF_INET6, prefixV6.lower.description, &p)
+        // `inet_pton` rejects a zone suffix, so render the network address on
+        // its own. A vmnet prefix is never link-scoped, so dropping the zone
+        // here loses nothing.
+        inet_pton(AF_INET6, IPv6Address(prefixV6.lower.value).description, &p)
```
jglogan: @vyncint 

```suggestion
        // `inet_pton` rejects a zone suffix, so render the network address on
        // its own. A vmnet prefix is never link-scoped, so dropping the zone
        // here loses nothing.
        guard inet_pton(AF_INET6, IPv6Address(prefixV6.lower.value).description, &p) == 0 else {
            throw ContainerizationError(...)
        }
```

I'd prefer we be strict here.

jglogan (changes requested):
@vyncint Great catch. Thanks for the fix.

For the VmnetNetwork API I'd prefer that we make the caller responsible for providing a zoneless address. Could you make that change?

Sources/Containerization/VmnetNetwork.swift
vyncint: Done in 9b9907d. `configurePrefixV6` no longer strips the zone: it renders `prefixV6.lower` as-is and guards on the `inet_pton` result, so a zoned prefix is rejected instead of quietly coerced, and supplying a zoneless prefix is the caller's job.

One deviation from the suggestion — `inet_pton` returns 1 on success and 0 when the string doesn't parse, so the guard reads `== 1`. As written (`== 0`) it would have thrown on every valid prefix.

Checking the result also closes something older than this PR: it was discarded, so any address `inet_pton` refused left `p` zeroed and configured `::`.

No call sites needed changing — every in-repo caller already passes a zoneless prefix (`CIDRv6("fd00::/64")` in the integration suite).

vyncint:
@jglogan pushed 9b9907d — the vmnet helper now leaves the zoneless prefix to the caller and checks `inet_pton`'s result instead of stripping the zone itself. Details in the [thread](https://github.com/apple/containerization/pull/827#discussion_r3732422769) (one note: the guard is `== 1`, since `inet_pton` returns 1 on success).

Re-verified locally: `make containerization` clean with `WARNINGS_AS_ERRORS=true`, 591 tests in 81 suites pass, format lint and `make check-licenses` clean. Ready for another look.

jglogan:
> Details in the [thread](https://github.com/apple/containerization/pull/827#discussion_r3732422769) (one note: the guard is `== 1`, since `inet_pton` returns 1 on success).

@vyncint Ah, thanks. I had a quick look at the manpage but just assumed 0 as success and negative as failure. Got it!

Looks like we've got GH actions again so it's building now.

jglogan (approved):

jglogan:
@vyncint Thank you, it's merged now!

PR #828, opened 2026-07-31
Use a custom volume for build cache inside of a container

adityaramani:
Keep in-container Swift build directories away from the virtiofs `/workspace` share and use a named volume instead for the cache store

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,14 @@
 # Build configuration variables
 BUILD_CONFIGURATION ?= debug
 WARNINGS_AS_ERRORS ?= true
-SWIFT_CONFIGURATION := $(if $(filter-out false,$(WARNINGS_AS_ERRORS)),-Xswiftc -warnings-as-errors) --disable-automatic-resolution
+
+# Allow for a custom build cache directory
+# By default this is left unset, and swift uses the default directory as `./.build`
+# The `linux_run` target exports SCRATCH_ROOT inside of the container
+SCRATCH_ROOT ?=
+SCRATCH_PATH ?= $(if $(SCRATCH_ROOT),$(SCRATCH_ROOT)/build-containerization)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,7 +15,14 @@
 # Build configuration variables
 BUILD_CONFIGURATION ?= debug
 WARNINGS_AS_ERRORS ?= true
-SWIFT_CONFIGURATION := $(if $(filter-out false,$(WARNINGS_AS_ERRORS)),-Xswiftc -warnings-as-errors) --disable-automatic-resolution
+
+# Allow for a custom build cache directory
+# By default this is left unset, and swift uses the default directory as `./.build`
+# The `linux_run` target exports SCRATCH_ROOT inside of the container
+SCRATCH_ROOT ?=
+SCRATCH_PATH ?= $(if $(SCRATCH_ROOT),$(SCRATCH_ROOT)/containerization)
```
jglogan: nit: perhaps the suffix should be `build-containerization` so the role of this is more obvious?

jglogan: not necessary for this PR but if it makes sense we could add it to any subsequent PR

Makefile
```diff
@@ -140,7 +162,7 @@ endif
 
 .PHONY: linux-test
 linux-test:
-	$(call linux_run,swift test $(SWIFT_CONFIGURATION))
+	$(call linux_run,swift test $(SWIFT_CONFIGURATION) --scratch-path $(LINUX_SCRATCH_ROOT)/containerization)
```
jglogan: Same name convention here `build-containerization`.

jglogan (approved):
lgtm, only nit is perhaps we should prefix any replacement for `.build` with `build-`?

jglogan (approved):

PR #831, opened 2026-08-05
Skip the default route when an interface has no gateway

### LINKED ISSUE #825: [Bug]: LinuxContainer installs a link-scope default route when ipv4Gateway is nil
### I have done the following

- [x] I reproduced the issue with Containerization 0.33.1 and confirmed the same logic remains on `main`

### Steps to reproduce

1. Create a custom type conforming to `Containerization.Interface`.
2. Supply an IPv4 address and return `nil` from `ipv4Gateway`:

```swift
struct GatewaylessInterface: Interface, VZInterface {
    let base: any Interface & VZInterface
    var ipv4Address: CIDRv4 { base.ipv4Address }
    var ipv4Gateway: IPv4Address? { nil }
    var macAddress: MACAddress? { base.macAddress }
    var mtu: UInt32 { base.mtu }
    func device() throws -> VZVirtioNetworkDeviceConfiguration { try base.device() }
}
```

3. Start a `LinuxContainer` whose first and only interface is this gatewayless interface.
4. Inside the guest, run:

```sh
ip -4 route show
ip route get 1.1.1.1
```

The issue is deterministic in a real Virtualization.framework Linux microVM.

The 0.33.1 public contract documents `ipv4Gateway` as “The IP address for the default route, or nil for no default route”:
https://github.com/apple/containerization/blob/0.33.1/Sources/Containerization/Interface.swift#L25-L26

However, `LinuxContainer` calls `routeAddDefault(name:ipv4Gateway:)` with `nil` for the first interface:
https://github.com/apple/containerization/blob/0.33.1/Sources/Containerization/LinuxContainer.swift#L631-L654

Current `main` retains the same semantic behavior through `setupInterface`: when both gateways are nil it logs “no gateway” and still calls `routeAddDefault`:
https://github.com/apple/containerization/blob/main/Sources/Containerization/VirtualMachineAgent%2BInterface.swift#L76-L82

### Current behavior

The guest receives an on-link default route even though `ipv4Gateway` is nil:

```text
default dev eth0 scope link
172.16.0.0/24 dev eth0 scope link src 172.16.0.3
```

Consequently, route lookup treats arbitrary external destinations as directly reachable on `eth0` and attempts neighbor resolution on that link.

This does not necessarily provide external connectivity when the backing attachment is private and has no uplink, but it contradicts the public `Interface` contract and prevents callers from creating a first interface with no default route.

### Expected behavior

When `ipv4Gateway` is nil, `LinuxContainer` should configure the IPv4 address and link route without installing an IPv4 default route. IPv4 and IPv6 default-route installation should follow their respective optional gateway values.

### Environment

- OS: macOS 26.5.2 (25F84)
- Xcode: 26.6 (17F113)
- Swift: Apple Swift 6.3.1
- Containerization: 0.33.1
- Architecture: Apple silicon / arm64

### Relevant log output

```text
default dev eth0 scope link
172.16.0.0/24 dev eth0 scope link src 172.16.0.3
```

### Code of Conduct

- [x] I agree to follow this projects Code of Conduct

adityasingh2400:
The `Interface` protocol documents `ipv4Gateway` and `ipv6Gateway` as the address for the default route, or nil for no default route, but `setupInterface` logged the nil-gateway case and then called `routeAddDefault` anyway. The guest ended up with `default dev eth0 scope link`, so route lookup treated arbitrary external destinations as directly reachable on the link and started neighbor resolution for them. This also left callers with no way to build a first interface that has no default route.

This returns early when both gateways are nil and leaves every other path untouched, so a v4-only gateway, a v6-only gateway, or a gateway outside the subnet all still install the same routes as before.

The new tests drive `setupInterface` through a recording `VirtualMachineAgent` double, which covers the routing decisions without booting a sandbox. `noDefaultRouteWhenBothGatewaysAreNil` fails on main and passes here, and `defaultRouteInstalledForIPv6OnlyGateway` pins the v6-only behavior that the existing integration test `testIPv6OnlyDefaultRoute` exercises against a real VM.

Fixes #825

### DIFF
--- a/Sources/Containerization/VirtualMachineAgent+Interface.swift
+++ b/Sources/Containerization/VirtualMachineAgent+Interface.swift
@@ -73,9 +73,14 @@ extension VirtualMachineAgent {
             )
         }
 
-        if ipv4Gateway == nil && ipv6Gateway == nil {
-            logger?.debug("no gateway for \(name)")
+        guard ipv4Gateway != nil || ipv6Gateway != nil else {
+            // `Interface` documents a nil gateway as "no default route", so
+            // installing one anyway would give the guest an on-link default
+            // route that makes every destination look directly reachable.
+            logger?.debug("no gateway for \(name), skipping the default route")
+            return
         }
+
         try await routeAddDefault(
             name: name,
             route: .init(ipv4Gateway: ipv4Gateway, ipv6Gateway: ipv6Gateway)
--- a/Tests/ContainerizationTests/InterfaceTests.swift
+++ b/Tests/ContainerizationTests/InterfaceTests.swift
@@ -14,7 +14,10 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
+import ContainerizationError
 import ContainerizationExtras
+import ContainerizationOCI
+import Foundation
 import Testing
 
 @testable import Containerization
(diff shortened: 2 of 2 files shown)


adityasingh2400:
One thing worth flagging for review: `vminitd`'s `ipRouteAddDefault` has a branch that deliberately installs a gatewayless v4 default route when the request carries no v4 gateway and no v6 gateway, commented as preserving pre-IPv6 behavior. I left that guest-side branch alone on purpose, so an older host talking to a newer `vminitd` keeps the exact behavior it has today and the wire contract does not change. The fix is host-side only, which is also where the `Interface` documentation lives.

Happy to move the check into the guest instead, or to gate it behind an explicit opt-in on `Interface`, if you would rather the compatibility path be retired at the same time.

PR #649, opened 2026-04-04
Add deep extent tree support (depth >= 2) in EXT4 reader

gbfish:
## Summary
- Refactor `getExtents()` to support extent trees of any depth via recursive descent
- Extract `readExtentNode()` helper that follows ExtentIndex nodes until reaching depth-0 ExtentLeaf nodes
- Consistent use of `.loadLittleEndian(as:)` for all on-disk struct reads
- Add test for multi-extent file readback

Fixes #645 (reader portion)

## Test plan
- [x] All 38 existing + new EXT4 tests pass
- [x] Depth 0 and depth 1 files read correctly (verified by existing test suite)
- [ ] Depth >= 2 will be fully testable after formatter support lands (PR 2)

Issue #645, opened 2026-04-03
[Request]: Implement deep extent tree support (depth >= 2) in ContainerizationEXT4


gbfish:
## Summary

The EXT4 reader and formatter currently only support extent trees of depth 0 and 1. When encountering depth >= 2, the reader throws `deepExtentsUnimplemented` and the formatter throws `fileTooBig`.

## Where in the code

- **Reader**: `EXT4+Reader.swift` line 246 — `getExtents()` has a `default` case that throws `Error.deepExtentsUnimplemented`
- **Formatter**: `EXT4+Formatter.swift` line 1168 — `writeExtents()` has a `default` case that throws `Error.fileTooBig`
- **Related bug**: `EXT4+Formatter.swift` line 1148 — force-unwrap `leafNode.leaves.last!.block` can crash if the leaves array is empty at a depth-1 boundary edge case

## When this matters

- Reading a pre-existing ext4 image (not created by this formatter) that contains files with highly fragmented extent trees (depth >= 2)
- Writing a file with more than ~1,360 non-contiguous extents (exceeds depth-1 capacity of 4 x 340 leaves)
- Overlay layer whiteouts that cause block fragmentation across multiple unpack passes

These are uncommon in typical container workflows (freshly written images are mostly contiguous), but the current behavior is a hard crash/error with no workaround.

## Proposed changes

1. **Reader**: Convert the depth-1 special case in `getExtents()` to a recursive/iterative descent that handles any depth — follow ExtentIndex nodes until reaching depth-0 ExtentLeaf nodes
2. **Formatter**: Extend `writeExtents()` to build multi-level extent trees when the number of extents exceeds depth-1 capacity
3. **Bug fix**: Replace the force-unwrap at line 1148 with safe optional access
4. **Tests**: Add test cases for reading and writing extent trees at depth 0, 1, and 2+

I'd like to work on this. Planning to split into two PRs:
- PR 1: Reader support for depth >= 2 + tests
- PR 2: Formatter support for depth >= 2 + force-unwrap fix + tests

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #462, opened 2025-12-30
Fix EXT4 extent offset double-counting causing block address overflow

Ronitsabhaya75:
Closes #441

## current behavior:
EXT4 filesystem corruption occurs when creating large files in containers, causing block addresses to exceed maximum inode values. This results in errors like:

```EXT4-fs warning (device vdb): ext4_block_to_path:105: block 1710731624 > max in inode 21209237```

### The fix:

Remove the pre-added offset. Connect to the correct block address immediately by passing blocks.start directly, allowing offset to be applied once inside ```fillExtents```. This ensures block addresses remain within filesystem bounds for files of any size.

PR #849, opened 2026-08-24
EXT4: guard extent-tree parsing against out-of-bounds reads on malformed images

iabdullah215:
## What

Bounds-check the extent-tree decode in `EXT4.EXT4Reader` so a malformed inode can no longer drive `Data.subdata` past the inode's block region. The decode is extracted into an internal `decodeExtents(inodeBlock:)` helper so it can be unit-tested directly.

## Why

An ext4 image is untrusted input. `getExtents` iterated `0..<entries` reading fixed-size records (12 bytes each) with `subdata`, trusting the entry counts from the on-disk extent headers without checking them against the bytes available:

- The inode block region is only 60 bytes (header + up to four 12-byte records). A depth-0 header claiming >4 leaves, or a depth-1 header claiming >4 indices, makes a read run past the block.
- In the depth-1 path, each leaf block trusts its own header's `entries` the same way, past the end of the block buffer.

Because `Data.subdata(in:)` has a precondition on its range, either case traps the process (`SIGTRAP`) instead of returning an error, a denial of service for any caller that reads a file backed by a crafted inode.

## Fix

- Guard every fixed-size read with `offset + size <= buffer.count` (throwing `EXT4.Error.invalidExtents` when a header claims more records than fit, i.e. surfacing the corruption rather than silently truncating).
- Fix a latent reversed range when reading a leaf block's header: `blockOffset..<extentHeaderSize` → `blockOffset..<blockOffset + extentHeaderSize` (previously benign only because `blockOffset == 0` there).

Well-formed images are unaffected: their headers never claim more records than the block holds.

## Testing

New `Tests/ContainerizationEXT4Tests/TestEXT4Reader+ExtentBounds.swift`:

- `wellFormedDepth0ExtentsParse` — regression: valid depth-0 extents still parse.
- `tooManyDepth0EntriesDoesNotTrap` — a header claiming 5 leaves in a 60-byte block is rejected (crashes the reader without this change).
- `shortInodeBlockReturnsEmpty` / `nonExtentInodeReturnsEmpty`, no-extent cases.

All existing `EXT4PathIOTests` (which read real files through `getExtents`) still pass.

## Related

- **Nearest misses:** #639 (*Fix a force-unwrap crash on the last leaf of an extent tree*) and #462 (*extent offset double-counting causing block address overflow*) hardened other extent-decode failures but did not bound the entry count against the buffer; this PR adds that check.
- Same class of malformed-image reader hardening: #663 (crash-instead-of-truncate on type overflow), and the directory-entry bounds fix in a sibling PR on the same reader.
- depth ≥ 2 remains an explicit `throw` (#645 / #649 / #650 track it as a feature); this PR doesn't change that.


iabdullah215:
@dkovba, mind giving these a look when you have a moment? #848 and #849 have had no CI since Aug 24; fork PRs need a maintainer to approve the workflow run. Both are bounds checks in EXT4+Reader.swift on the untrusted-image parse path, same shape as the xattr loop bounds you fixed in #658: a malformed inode's entries count drives subdata past the end of the block and traps the reader. Tests included in each.

PR #850, opened 2026-08-25
Add Pi agent support to sandboxy

### LINKED ISSUE #801: Add Pi support to sandboxy
Hi, I’d like to propose adding Pi as a built-in agent to the experimental `sandboxy` example.

I’ve been using sandboxy to run coding agents and got Pi working with the PTY support that is already in place. Pi’s interactive UI needs a real terminal; once it is launched through the existing PTY path, it works well.

The proposed change is small:

- Add Pi’s npm package and `pi` launch command.
- Persist Pi’s agent configuration from `~/.pi/agent`.
- Forward the common provider credentials.
- Allow the provider and npm hosts through the existing proxy.
- Avoid applying Claude’s `global-agent` preload to Pi, since Pi does not install that package.

This does not change the VM, networking, or sandboxing model. Pi uses the same VM isolation, host-only network, filtered proxy, and existing PTY relay as the other agents. The install uses npm’s `--ignore-scripts` option.

I have a small local implementation and have tested Pi interactively on macOS 26. Before opening a PR, I wanted to check whether this is a direction the maintainers would be interested in for sandboxy, or whether you would prefer agent definitions to remain user-provided.

AntApper:
Resolves #801

### Overview
Adds built-in agent support for Pi (`@earendil-works/pi-coding-agent`) to `sandboxy`. This serves as a second built-in agent definition alongside Claude Code, validating the multi-agent application architecture.

### Changes
- Adds `AgentDefinition.pi` configured with:
  - Base image `docker.io/library/node:22`
  - NPM installation of `@earendil-works/pi-coding-agent` with `--ignore-scripts`
  - Mount for `~/.pi/agent` to `/root/.pi/agent`
  - Automatic forwarding for provider credentials (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GEMINI_API_KEY`, `GOOGLE_API_KEY`, `OPENROUTER_API_KEY`, `XAI_API_KEY`, `ANTHROPIC_OAUTH_TOKEN`)
  - Allowed host domains for provider endpoints and npm
- Conditions `global-agent` proxy bootstrapping in `RunAgentCommand` so it is only applied when `global-agent` is present in the agent's install commands.
- Cleans up `sigwinchStream` on exit in `runContainerSession`.
- Updates `sandboxy` README documentation and usage examples.

### Verification
- Built `sandboxy` binary and verified clean compilation (`make -C examples/sandboxy`).
- Ran code formatting (`make fmt`) and license/linter checks (`make check`).
- Ran the full test suite (`swift test`, 593 tests passed).
- Verified `sandboxy config list --agents` and interactive launch with PTY relay.

### DIFF
--- a/examples/sandboxy/Sources/sandboxy/RunAgentCommand.swift
+++ b/examples/sandboxy/Sources/sandboxy/RunAgentCommand.swift
@@ -687,6 +687,7 @@ private func runContainerSession(
         envVarsBuilder.append("https_proxy=\(proxyURL)")
         envVarsBuilder.append("NO_PROXY=localhost,127.0.0.1")
         envVarsBuilder.append("no_proxy=localhost,127.0.0.1")
+        // TODO: Refine install customization as we integrate more agent definitions rather than inspecting installCommands directly.
         if definition.installCommands.contains(where: { $0.contains("global-agent") }) {
             envVarsBuilder.append("GLOBAL_AGENT_HTTP_PROXY=\(proxyURL)")
             envVarsBuilder.append("GLOBAL_AGENT_HTTPS_PROXY=\(proxyURL)")
--- a/examples/sandboxy/Sources/sandboxy/RunAgentCommand.swift
+++ b/examples/sandboxy/Sources/sandboxy/RunAgentCommand.swift
@@ -654,7 +654,10 @@ private func runContainerSession(
     let sigwinchStream = AsyncSignalHandler.create(notify: [SIGWINCH])
     let current = try Terminal.current
     try current.setraw()
-    defer { current.tryReset() }
+    defer {
+        sigwinchStream.cancel()
+        current.tryReset()
+    }
 
     // Build environment for the agent process.
     var envVarsBuilder = [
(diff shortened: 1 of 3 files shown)


examples/sandboxy/Sources/sandboxy/RunAgentCommand.swift
```diff
@@ -684,17 +687,17 @@ private func runContainerSession(
...
-        envVarsBuilder.append("GLOBAL_AGENT_NO_PROXY=localhost,127.0.0.1")
-
-        // Prepend global-agent bootstrap to NODE_OPTIONS so Node.js http/https
-        // modules respect the proxy environment variables.
-        if let idx = envVarsBuilder.firstIndex(where: { $0.hasPrefix("NODE_OPTIONS=") }) {
-            let existing = String(envVarsBuilder[idx].dropFirst("NODE_OPTIONS=".count))
-            envVarsBuilder[idx] = "NODE_OPTIONS=-r /usr/local/lib/node_modules/global-agent/dist/routines/bootstrap.js \(existing)"
-        } else {
-            envVarsBuilder.append("NODE_OPTIONS=-r /usr/local/lib/node_modules/global-agent/dist/routines/bootstrap.js")
+        if definition.installCommands.contains(where: { $0.contains("global-agent") }) {
```
jglogan, line 691: @AntApper I don't have a deep understanding of what's necessary when embedding agents in containers but having all this mostly-nodes-related stuff conditionalized here feels somewhat arbitrary.

Can you think of any way to do this declaratively as part of the agent definition, such that it's secure (for example, doesn't allow an agent setup to perform arbitrary operations on the host or access user data)?

jglogan: Let's use the conversation here to discuss approaches before making any code changes.

AntApper: completely agree, sniffing the install commands was a bit of a hack and having node paths hardcoded in the runner is messy. 

the only reason it's there is claude code doesn't respect standard HTTP_PROXY env vars without global-agent, whereas pi doesn't need it (and node crashes if it tries to require a file that isn't installed).

to make it declarative and safe from arbitrary user JSONs, we could do a simple enum like `proxyBootstrap: .standard | .nodeGlobalAgent` so it's strictly bounded, or something like a `proxyEnv` dictionary on the agent definition with a `{PROXY_URL}` token so sandboxy doesn't need any node-specific logic at all. since it's just setting container env vars either way, host files/execution wouldn't be exposed.

let me know if either of those sounds good to you or if you were picturing something different!

jglogan: @AntApper I was picturing something more generic but without going through the exercise of looking at what it'd take to support a variety (more than just n=2) of agents, there's the risk of building the wrong thing.

Could you just put a TODO comment above this block that notes that there's some smell here, and that we should look at refining the install customization as we integrate more agent definitions?

AntApper: Done! Added the TODO comment above the check.

jglogan (approved):

jglogan:
@AntApper merged, thank you for your contribution!

Issue #883, opened 2026-08-27
The NBD export does not honour command flags, read-only, or bounds, and cannot keep its blocks anywhere but a file


MayCXC:
Two related things, one a set of protocol defects in the NBD server this library ships, the other a use for it.

## The NBD export does not hold to what it tells a client

Reading `Sources/.../NBD` against the protocol document turned up five places where the server tells a client one thing and does another. https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md

- **The command flags are read past.** They sit in the two bytes after the request magic, and are skipped, so an export that advertises force unit access replies to a command carrying it before what it wrote is durable. A command with that flag has to wait.
- **A read-only export does not turn writes away**, which the protocol requires of it. The pod volume API already offers to ask for read-only, so a caller can ask today and not get it.
- **A request reaching past the end is not refused as invalid.**
- **Structured replies are not offered.** A client that asks for replies carrying their own framing does not get them, and that framing is what lets an export be asked how it is laid out: a read answered that way names the offset it covers and carries a message with any error rather than a bare number.
- **`NBD_OPT_LIST` does not name the export on offer**, and flush is not stated to cover what other connections wrote, which is the condition for telling a client it may spread its work across several connections. Every connection here serves the one store behind the export, so the guarantee holds and is simply not stated.

These are worth fixing whether or not anything below lands, so I can send them as their own pull request.

## Letting an export keep its blocks somewhere other than a file

The export writes through to a file. Letting the store behind it be chosen makes a memory-backed store possible, which is what I wanted it for: swapping a container to an area held in host memory rather than on disk. With the block-descriptor context above, a store in memory knows which chunks it holds and a file is asked through the same seeks a sparse copy uses, so the holes a client is told about are the ones really there.

## Testing note

The memory-backed test asserts the release as well as the high water mark: the filler exits while the guest is still up, freeing its swap slots, and a freed cluster is discarded, so the export drops to little more than the swap header. The discards trail the exit, so the test watches the export while the machine is still up rather than sampling it once. A test that samples after shutdown measures nothing.

## Question

Is a memory-backed swap area something you want in the library, or is the useful half here just the protocol fixes plus the pluggable store? I am happy to send the first two sections alone.

Depends on the swap work in #877.

PR #895, opened 2026-08-27
Hold the NBD export to what it tells a client, and let it keep its blocks somewhere other than a file

MayCXC:
## Summary

Five places where the NBD server tells a client one thing and does another, read against the protocol document, plus a pluggable block store and one use for it. https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md

**The command flags were read past.** They sit in the two bytes after the request magic, which were skipped, so an export that said it took force unit access replied to a command carrying it before what it wrote was durable. They are read now, and a command carrying that flag waits.

**A read-only export now says so and turns writes away,** which the protocol requires of it and which the pod volume API already offers to ask for. A request reaching past the end is refused as invalid rather than served.

**Structured replies are offered and answered.** A client that asks for replies carrying their own framing gets them, and a read answered that way names the offset it covers and carries a message with any error rather than a bare number. That framing is what lets an export be asked how it is laid out, so the block-descriptor context is offered and answered: a store in memory knows which chunks it holds and a file is asked through the same seeks a sparse copy uses, so the holes a client is told about are the ones really there.

**Listing the exports names the one on offer.** Every connection serves the one store behind the export, so a flush on any of them covers what was written on the others, which is the condition for saying that a client may spread its work across several.

**The store behind an export is chosen rather than fixed to a file,** which is what makes a memory-backed store possible: a container's swap area held in host memory rather than on disk.

## Dependency and Merge Order

This targets stock `apple/containerization:main`. **Merge order: #863, #872, #878, then this.** The head carries their commits because all of them target stock `main`; after they merge, this branch takes current `main` and its remaining diff is the six commits above.

## Motivation and Context

Discussion: #883, where I ask whether the memory-backed swap area is something you want at all, or whether the useful half here is the protocol fixes plus the pluggable store. The protocol fixes stand on their own and I am glad to send them separately.

## Testing

- `make containerization`, `make check`, and `swift test` clean.
- `make integration`: the memory-backed test asserts the release as well as the high water mark. The filler exits while the guest is still up, freeing its swap slots, and a freed cluster is discarded, so the export drops to little more than the swap header. The discards trail the exit, so the test watches the export while the machine is still up rather than sampling it once; sampling after shutdown measures nothing.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ LIBARCHIVE_UPSTREAM_REPO := https://github.com/libarchive/libarchive
 LIBARCHIVE_UPSTREAM_VERSION := v3.7.7
 LIBARCHIVE_LOCAL_DIR := workdir/libarchive
 
-KATA_BINARY_PACKAGE := https://github.com/kata-containers/kata-containers/releases/download/3.17.0/kata-static-3.17.0-arm64.tar.xz
+KATA_BINARY_PACKAGE := https://github.com/kata-containers/kata-containers/releases/download/3.28.0/kata-static-3.28.0-arm64.tar.zst
 CLOUD_HYPERVISOR_URL := https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v52.0/cloud-hypervisor-static-aarch64
 # SHA256 of the v52.0 aarch64 static binary (verified locally from the
 # upstream release artifact). Bump alongside CLOUD_HYPERVISOR_URL.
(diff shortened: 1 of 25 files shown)


crosbymichael:
Thank you for your interest in contributing. There are real people on the other side of the screen that must review these changes and the changes of others in the community. 

Opening 25+ PRs at the same time, some over 1000+ lines of change and many with dependencies on one another, makes meaningful review impossible. You as a contributor are responsible for the dependency management and order that you open PRs. Open the base PRs, get them merged, then open the follow-up PRs. Significant changes should also be proposed first by opening an issue and engaging with maintainers.

We're closing these PRs. If you're still interested in contributing changes to this and other projects, please review our CONTRIBUTING.md and AI contribution policy. Start by opening an issue to discuss your proposed changes and ensure they address project needs, then submit tightly scoped PRs 1–2 at a time. This allows us to give you and other contributors the proper time and meaningful feedback.

PR #903, opened 2026-09-01
Add logic to cctl to enable NBD mounts

noah-thor:
This enables cctl to handle NBD mounts for formatted and unformatted NBD mounts into Linux containers.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -95,6 +95,14 @@ let package = Package(
             ]
         ),
         .testTarget(
+            name: "cctlTests",
+            dependencies: [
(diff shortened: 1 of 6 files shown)


crosbymichael:
Are the new tests you added the only time we have tested `cctl` in the suite?

noah-thor:
> Are the new tests you added the only time we have tested `cctl` in the suite?

As far as I know yes, I did not see any other tests in the integration suites driving the `cctl` binary.

crosbymichael (approved):

PR #906, opened 2026-09-02
Update cctl to respect entrypoint and cmd

noah-thor:
This expands `cctl run` to utilize entrypoint and cmd when present in the OCI image, instead of always defaulting to `/bin/sh`. The user can still provide explicit commands if necessary to override these values.

### DIFF
--- a/Sources/cctl/RunCommand.swift
+++ b/Sources/cctl/RunCommand.swift
@@ -99,6 +99,16 @@ extension Application {
         @Option(name: .long, help: "Current working directory")
         var cwd: String = "/"
 
+        @Option(
+            name: .customLong("entrypoint"),
+            help: """
+                Override the image's ENTRYPOINT with a single executable \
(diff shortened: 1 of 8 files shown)


Sources/cctl/RunCommand.swift
```diff
@@ -245,6 +255,22 @@ extension Application {
             ).first!
             .appendingPathComponent("com.apple.containerization")
         }()
+
+        /// Resolve the command from the image config. The guest is Linux
+        /// regardless of the host, so the config is read for linux/arm64 rather
+        /// than `Platform.current`.
```
noah-thor: This comment is confusing, will re-word. It is indicating that `cctl run` always uses the image config for `Platform.current`. 

https://github.com/apple/containerization/blob/main/Sources/Containerization/ContainerManager.swift#L311

katiewasnothere (approved):

PR #832, opened 2026-08-06
fix: memory over allocation due to last partial block group

siddh34:
## Description

This resolves issue #647 

This is fix for memory over allocation due to last partial block

I have explicitly laid out metadata for extra/last groups

## Modifications: 

Packed-region math is computation starts at line [718](https://github.com/siddh34/containerization/blob/98555ce9ba67477bb2c92569639809fbda938d59/Sources/ContainerizationEXT4/EXT4%2BFormatter.swift#L718) 

function to calculate blocks in last group at line [58](https://github.com/siddh34/containerization/blob/98555ce9ba67477bb2c92569639809fbda938d59/Sources/ContainerizationEXT4/EXT4%2BFormatter.swift#L58)

while loop here didn't made any sense so replaced it with if check line [914](https://github.com/siddh34/containerization/blob/98555ce9ba67477bb2c92569639809fbda938d59/Sources/ContainerizationEXT4/EXT4%2BFormatter.swift#L914)

PR #925, opened 2026-09-12
Fix/support for last partial group

siddh34:
@dkovba splitting #832

### DIFF
--- a/Sources/ContainerizationEXT4/EXT4+Formatter.swift
+++ b/Sources/ContainerizationEXT4/EXT4+Formatter.swift
@@ -612,6 +612,28 @@ extension EXT4 {
             throw Error.unsupportedFiletype
         }
 
+        private func markAllocatedRange(
+            start: UInt64,
+            end: UInt64,
+            groupStart: UInt64,
+            groupEnd: UInt64,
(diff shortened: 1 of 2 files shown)


Sources/ContainerizationEXT4/EXT4+Formatter.swift
```diff
@@ -96,7 +96,9 @@ extension EXT4 {
             guard blockSize >= 1024 && blockSize <= 4096 && blockSize.nonzeroBitCount == 1 else {
                 throw Error.invalidBlockSize(blockSize)
             }
-            guard minDiskSize / UInt64(blockSize) <= UInt64(UInt32.max) else {
+            let minimumBlockCount =
+                minDiskSize / UInt64(blockSize) + (minDiskSize.isMultiple(of: UInt64(blockSize)) ? 0 : 1)
```
JaewonHur: Can we do?
```
let minimumBlockCount = (minDiskSize - 1) / UInt64(blockSize) + 1
```

siddh34: done

JaewonHur (approved):

PR #926, opened 2026-09-15
vminitd: make the runc-backed launch path work, and add a default sec…

crosbymichael:
…comp profile

`LinuxContainer.Configuration.ociRuntimePath` has existed on main for some time and looks complete: the field is plumbed through the proto, and `ManagedContainer` already branched on it to build a `RuncProcess`. It has never worked. Nothing ever placed a runc binary inside the guest, and once one is there, `runc create` deadlocks on its own output capture before any container starts.

This makes it work, and then uses it for what it is for: runc implements seccomp and the advanced LSMs, so they do not have to be reimplemented in Swift. A pure-Swift seccomp compiler was evaluated and abandoned; delegating to runc is the alternative.


Sources/Containerization/LinuxContainer.swift
```diff
@@ -375,6 +428,18 @@ public final class LinuxContainer: Container, Sendable {
         self.writableLayer = writableLayer
     }
 
+    /// Refuse a seccomp profile that no runtime will install. `vmexec` does not
+    /// read `spec.linux.seccomp`, so such a container would run unfiltered
+    /// while every observable said it was sandboxed.
+    private static func requireOCIRuntimeForSeccomp(_ configuration: Configuration) throws {
+        guard configuration.ociRuntimePath != nil else {
```
adityaramani, line 435: if someone specified `ociRuntimePath = /sbin/vmexec` this would not throw but the seccomps would still not be supported right?

Sources/Containerization/LinuxContainer.swift
```diff
@@ -507,6 +596,59 @@ public final class LinuxContainer: Container, Sendable {
...
+    /// ``defaultMounts()`` mounts `/dev` as devtmpfs, which `vmexec` is happy
+    /// with. devtmpfs is a single kernel-wide instance, so runc's
+    /// `mountConsole()` cannot create `/dev/console` there (EPERM) and
+    /// `process.terminal = true` fails before the console socket is contacted.
+    ///
+    /// Only the stock `/dev` entry is rewritten, and only when
+    /// ``Configuration/ociRuntimePath`` is set. Runs at mount-assembly time, so
+    /// the order a configuration closure sets its properties in doesn't matter.
+    private func mountsForRuntime() -> [Mount] {
+        guard self.config.ociRuntimePath != nil else {
```
adityaramani, line 625: similar as above? Should the check be `if  ! vmexec`

vminitd/Sources/VminitdCore/ContainerProcess.swift
```diff
@@ -29,7 +29,10 @@ struct ContainerExitStatus: Sendable {
 ///
 /// This protocol abstracts the underlying container runtime implementation,
 /// allowing for different backends like vmexec or runc.
-protocol ContainerProcess: Sendable {
+///
+/// Class-bound so owners can track a process by identity: `pid` is `nil` before
+/// start and, for the runc implementations, `nil` again after exit.
+protocol ContainerProcess: AnyObject, Sendable {
```
adityaramani, line 35: Why do we need `AnyObject`?

crosbymichael: It makes comparisons work.  We could handle this a different way if it end up to be an issue.

Sources/Containerization/LinuxContainer.swift
crosbymichael: correct. We can consider vmexec in the ociRuntimePath a misuse.  Added some additional checks.

adityaramani (approved):

PR #930, opened 2026-09-16
Add filesystem stats in ContainerStatistics

JaewonHur:
* Add `filesystem (block_size, blocks, free_blocks, inodes, free_inodes)` field to protobuf `ContainerStats`.
  * Renamed `files`, `ffree` to `inodes` and `free_inodes` as those are POSIX's abstract.
* Add `ManagedContainer.filesystemStats` which stats the file system using `statfs` syscall. This exposes all fields in `statfs` struct.

* Update `ContainerStatistics` to include `filesystem` fields.
  * Caller can compute `usedBytes = (blocks - freeBlocks) * blockSize`, `inodesUsed = inodes - freeInodes`.
  * Other numbers can be computed if needed (e.g., `FS capacity = blocks * blockSize`).

* Added `CZ_statfs` in `LCShim/syscall.c` instead of using `musl` or `glibc` because it fails to import `statfs` when compiling for x86_64 linux.


adityaramani (approved):
I think this looks good. Will want another reviewer to also +1 though

Sources/Containerization/SandboxContext/SandboxContext.pb.swift
```diff
@@ -1595,6 +1599,15 @@ public nonisolated struct Com_Apple_Containerization_Sandbox_V3_ContainerStats:
...
 
+  public var filesystem: Com_Apple_Containerization_Sandbox_V3_FilesystemStats {
+    get {_storage._filesystem ?? Com_Apple_Containerization_Sandbox_V3_FilesystemStats()}
+    set {_uniqueStorage()._filesystem = newValue}
+  }
+  /// Returns true if `filesystem` has been explicitly set.
+  public var hasFilesystem: Bool {_storage._filesystem != nil}
+  /// Clears the value of `filesystem`. Subsequent reads from it will return its default value.
+  public mutating func clearFilesystem() {_uniqueStorage()._filesystem = nil}
+
```
noah-thor: Is this all generated via `protoc`?

Sources/Containerization/SandboxContext/SandboxContext.proto
```diff
@@ -427,6 +428,7 @@ message ContainerStats {
   BlockIOStats block_io = 5;
   repeated NetworkStats networks = 6;
   MemoryEventStats memory_events = 7;
+  FilesystemStats filesystem = 8;
```
noah-thor: Is there any sort of concept around multiple filesystems? For example `rootfs` + `nbd`

Sources/Containerization/ContainerStatistics.swift
```diff
@@ -242,7 +256,9 @@ public struct StatCategory: OptionSet, Sendable {
     public static let network = StatCategory(rawValue: 1 << 4)
     /// Memory event counters (OOM kills, pressure events, etc.).
     public static let memoryEvents = StatCategory(rawValue: 1 << 5)
+    /// Filesystem occupancy statistics.
+    public static let filesystem = StatCategory(rawValue: 1 << 6)
```
noah-thor, line 274: What does this `1 << 6` mean

vminitd/Sources/LCShim/include/syscall.h
```diff
@@ -104,4 +104,18 @@ int CZ_pidfd_getfd(int pidfd, int targetfd, unsigned int flags);
 
 int CZ_prctl_set_no_new_privs();
 
+// Mirrors the fields of struct statfs needed to compute filesystem usage.
+// struct statfs itself isn't reliably importable from Swift across all
+// target architectures via the Musl/Glibc modulemaps, so this wraps
```
noah-thor: Why is this the case?

vminitd/Sources/LCShim/syscall.c
```diff
@@ -49,4 +49,17 @@ int CZ_setrlimit(int resource, unsigned long long soft,
...
+  struct statfs s;
+  if (statfs(path, &s) != 0) {
+    return -1;
+  }
+  out->f_blocks = (unsigned long long)s.f_blocks;
+  out->f_bfree = (unsigned long long)s.f_bfree;
+  out->f_bsize = (unsigned long long)s.f_bsize;
+  out->f_files = (unsigned long long)s.f_files;
+  out->f_ffree = (unsigned long long)s.f_ffree;
+  return 0;
```
noah-thor, line 70: Is there a reason not to pass it directly in the `statfs` call by reference like: `statfs(path, out)`?

noah-thor (approved):

Sources/Containerization/SandboxContext/SandboxContext.pb.swift
JaewonHur: Yeah. You can run `make proto`.

Sources/Containerization/SandboxContext/SandboxContext.proto
crosbymichael: It maybe possible, good callout.

Sources/Containerization/ContainerStatistics.swift
JaewonHur: Followed the convention. Internally, the guest side checks that bit to stat filesystem or not.

JaewonHur: Though it'd be better using a named variable instead of a number..

noah-thor: Yeah I agree variable is better than magic number. I was just trying other understand why these `StatCategories` are bit based type not an enum or something else

vminitd/Sources/LCShim/include/syscall.h
JaewonHur: glibc's struct statfs (bits/statfs.h) branches its field types based on a preprocessor macro:
```
struct statfs {
    __fsword_t f_type;
    __fsword_t f_bsize;
#ifndef __USE_FILE_OFFSET64
    __fsblkcnt_t f_blocks; ...      // 32-bit variant
#else
    __fsblkcnt64_t f_blocks; ...    // 64-bit variant
#endif
    ...
};
```

__USE_FILE_OFFSET64 is controlled by _FILE_OFFSET_BITS at compile time — the same header name struct statfs describes two different memory layouts depending on what macros the including translation unit had set. There's also a second name, struct statfs64, gated behind a different macro (__USE_LARGEFILE64).

musl's struct statfs has no such branching — musl never had a 32-bit legacy off_t/blkcnt_t story, so there's exactly one unconditional layout, and statfs64/fstatfs64 are just #define aliases to the same functions, not a different struct.

That's the discrepancy: a Clang module (which is what a modulemap builds) is precompiled once, with one fixed macro configuration baked into its PCM. A type whose layout depends on _FILE_OFFSET_BITS isn't safe to modularize that way — whichever macro state the module happened to be built with becomes "the" struct statfs for every importer, silently disagreeing with any C code compiled under the other setting. musl's single, macro-independent layout has no such hazard, so its modulemap can freely declare sys_statfs/sys_vfs as real submodules. glibc's toolchain-maintained glibc.modulemap most likely just never added sys/vfs.h/sys/statfs.h for exactly this reason (or nobody has patched it in, since these modulemaps are hand-curated, not auto-generated from every header).

crosbymichael (approved):

PR #933, opened 2026-09-17
Fix socket forwarding hangs and shared I/O stalls

realrajaryan:
This PR addresses the SSH-agent forwarding hangs and terminal/exec stalls reported in [apple/container#2247](https://github.com/apple/container/issues/2247). The relay now registers both ends before forwarding starts and resumes blocked writes when the destination is ready, allowing other work on the shared I/O loop to continue. When connections close, buffered data is drained before EOF and stale events are discarded as file descriptors are reused. On the host, failed socket connections are closed individually so the listener keeps accepting new clients.


vminitd/Sources/VminitdCore/IOPair.swift
```diff
@@ -102,86 +140,55 @@ final class IOPair: Sendable {
...
+                let readFd = fcntl(io.from.fileDescriptor, F_DUPFD_CLOEXEC, 0)
+                guard readFd >= 0 else { throw POSIXError.fromErrno() }
+                io.readFd = readFd
+                let writeFd = fcntl(io.to.fileDescriptor, F_DUPFD_CLOEXEC, 0)
+                guard writeFd >= 0 else { throw POSIXError.fromErrno() }
+                io.writeFd = writeFd
+
+                try ProcessSupervisor.default.registerFd(writeFd, mask: .output) { mask in
+                    self.io.withLock { io in
+                        if mask.isHangup {
```
dkovba: Is a check for `ignoreHup` missing here?

```suggestion
                        if mask.isHangup && !ignoreHup {
```

Sources/Containerization/UnixSocketRelay.swift
```diff
@@ -138,18 +138,19 @@ extension UnixSocketRelay {
         state.withLock {
             $0.listener = listener
             $0.t = Task {
-                do {
-                    defer { try? listener.finish() }
-                    for await connection in listener {
+                defer { try? listener.finish() }
+                for await connection in listener {
+                    do {
```
dkovba, line 149: Should we add a similar `do-catch` block for `try await self.handleHostUnixConn()`?

Sources/Containerization/UnixSocketRelay.swift
```diff
@@ -199,15 +200,15 @@ extension UnixSocketRelay {
                 "hostFd": "\(hostSocket.fileDescriptor)",
                 "guestFd": "\(vsockConn.fileDescriptor)",
             ])
-        try hostSocket.connect()
-
         do {
```
dkovba, line 209: Should we use a similar `do-catch` block in `try await self.relay(hostConn: hostConn, ...)`?

dkovba (approved):
👍

